PR #1, opened 2022-08-05
Remove polyChkTable from bg collision check

krm01:
So this was a MASSIVE performance improvement in my project. I'm talking having 2 enemies in a room at 10 fps, to 7 enemies at 20 fps, in a large dungeon scene that was nearing the collision poly limit.

I suspect this was one of the original devs ideas of an optimization; don't check polys you've already checked for that update cycle, makes sense. but the data structure they used (array) just destroys any gains you would have gotten. It's significantly faster to just re-check the polys than it is to go through and clear the table after you're done checking, because that table gets large for dungeon scenes. And this is called through BgCheck_CheckLineImpl which is like the most foundational collision function in the whole BgCheck system, so it is called tons of times every frame. I am not kidding when I say I doubled my framerate and doubled the # of enemies loaded to maintain a consistent 20fps, just by ripping out this system.

### DIFF
--- a/src/code/z_bgcheck.c
+++ b/src/code/z_bgcheck.c
@@ -901,7 +901,7 @@ s32 BgCheck_CheckStaticCeiling(StaticLookup* lookup, u16 xpFlags, CollisionConte
 }
 
 /**
- * Tests if line `posA` to `posB` intersects with a static poly in list `ssList`. Uses polyCheckTbl
+ * Tests if line `posA` to `posB` intersects with a static poly in list `ssList`.
  * returns true if such a poly exists, else false
  * `outPoly` returns the pointer of the poly intersected
  * `posB` and `outPos` returns the point of intersection with `outPoly`
@@ -911,7 +911,6 @@ s32 BgCheck_CheckLineAgainstSSList(SSList* ssList, CollisionContext* colCtx, u16
                                    Vec3f* posB, Vec3f* outPos, CollisionPoly** outPoly, f32* outDistSq, f32 chkDist,
                                    s32 bccFlags) {
     SSNode* curNode;
-    u8* checkedPoly;
     Vec3f polyIntersect;
     CollisionPoly* polyList;
     CollisionPoly* curPoly;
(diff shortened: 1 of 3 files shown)


Zelllll:
Looks good to me, but how does this affect overworld maps? Does it also have the same gains? I know that overworld maps have different collision properties than most of the dungeons, so will this have any effect on that?

krm01:
So I think it will be a less noticeable improvement but it at least shouldn't be worse... the gains are more noticeable the more collision polys you have in the scene. I think I can do some more testing on this later with different scene types.

src/code/z_bgcheck.c
```diff
@@ -929,9 +928,8 @@ s32 BgCheck_CheckLineAgainstSSList(SSList* ssList, CollisionContext* colCtx, u16
     curNode = &colCtx->polyNodes.tbl[ssList->head];
     while (true) {
         polyId = curNode->polyId;
-        checkedPoly = &colCtx->polyNodes.polyCheckTbl[polyId];
 
-        if (*checkedPoly == true || COLPOLY_VIA_FLAG_TEST(polyList[polyId].flags_vIA, xpFlags1) ||
+        if (COLPOLY_VIA_FLAG_TEST(polyList[polyId].flags_vIA, xpFlags1) ||
             !(xpFlags2 == 0 || COLPOLY_VIA_FLAG_TEST(polyList[polyId].flags_vIA, xpFlags2))) {
 
             if (curNode->next == SS_NULL) {
```
Yanis42, line 935: Not one of your changes but this if statement looks better like that imo:
```c
if (curNode->next != SS_NULL) {
    curNode = &colCtx->polyNodes.tbl[curNode->next];
} else {
    break;
}
```
it shouldn't change anything afaik

I don't have anything else to say, really cool to have performance improvements 😄

Yanis42: just in case it's not clear I'm speaking of this if:
```c
if (curNode->next == SS_NULL) {
    break;
} else {
    curNode = &colCtx->polyNodes.tbl[curNode->next];
    continue;
}
```

Zelllll (approved):
Good to merge imo

PR #2, opened 2022-08-05
Some QOL improvements

Recardo-7:


Yanis42:
For the fast bunny hood I have this made from the rando ASM hack, not sure if the extra if I have is really necessary tho: https://github.com/Yanis42/oot/blob/68eb30a4c3ad86515fc0cf78776693e8e64b7564/src/overlays/actors/ovl_player_actor/z_player.c#L5950 (code given to me by Ticamus)

Recardo-7:
> For the fast bunny hood I have this made from the rando ASM hack, not sure if the extra if I have is really necessary tho: https://github.com/Yanis42/oot/blob/68eb30a4c3ad86515fc0cf78776693e8e64b7564/src/overlays/actors/ovl_player_actor/z_player.c#L5950 (code given to me by Ticamus)

I'm not sure exactly what this does, but it seems to slow down rolling and jumping, which I don't really see as useful. The player should be faster no matter how he moves.

Yanis42:
> > For the fast bunny hood I have this made from the rando ASM hack, not sure if the extra if I have is really necessary tho: https://github.com/Yanis42/oot/blob/68eb30a4c3ad86515fc0cf78776693e8e64b7564/src/overlays/actors/ovl_player_actor/z_player.c#L5950 (code given to me by Ticamus)
> 
> I'm not sure exactly what this does, but it seems to slow down rolling and jumping, which I don't really see as useful. The player should be faster no matter how he moves.
> 

Yeah I just wanted to share this just in case it was something useful

src/code/fault.c
```diff
@@ -1206,7 +1202,6 @@ void Fault_ThreadEntry(void* arg) {
         } else {
             // Draw error bar signifying the crash screen is available
             Fault_DrawCornerRec(GPACK_RGBA5551(255, 0, 0, 1));
-            Fault_WaitForButtonCombo();
```
Zelllll, line 1209: If you are going to remove the call here entirely, then you should also remove the function itself.

src/code/z_lifemeter.c
```diff
@@ -508,7 +510,11 @@ void Health_UpdateBeatingHeart(PlayState* play) {
             interfaceCtx->beatingHeartOscillatorDirection = 0;
             if (!Player_InCsMode(play) && (play->pauseCtx.state == 0) && (play->pauseCtx.debugState == 0) &&
                 Health_IsCritical() && !Play_InCsMode(play)) {
+                #ifdef NO_LOW_HEALTH_BEEP
+                Actor_SetColorFilter(&GET_PLAYER(play)->actor, 0x4000, 255, 0, 6);
```
Zelllll, line 514: I'd rather there be a separate ifdef for this new behavior of the player being red, but it may be better to not do this at all for now until we can discuss it further.

src/code/z_map_exp.c
```diff
@@ -364,7 +364,7 @@ void Minimap_Draw(PlayState* play) {
 
     OPEN_DISPS(play->state.gfxCtx, "../z_map_exp.c", 626);
 
-    if (play->pauseCtx.state < 4) {
+    if (play->pauseCtx.state < 3) {
```
Zelllll, line 367: What is the rationale for this change here?

src/code/z_parameter.c
```diff
@@ -3100,7 +3102,40 @@ void Interface_Draw(PlayState* play) {
         Gfx_SetupDL_39Overlay(play->state.gfxCtx);
 
         // Rupee Icon
+        #ifndef WALLET_COLORS
```
Zelllll, line 3105: I think the implementation here could be a little cleaner. Try this instead:
```
        // Rupee Icon
#ifndef WALLET_COLORS
        gDPSetPrimColor(OVERLAY_DISP++, 0, 0, 200, 255, 100, interfaceCtx->magicAlpha);
#else
        static Color_RGB8 const walletColors[] = {
            { 200, 255, 100 },
            { 130, 130, 255 },
            { 255, 100, 100 },
        };
        u8 walletUpg = CUR_UPG_VALUE(UPG_WALLET);
        gDPSetPrimColor(OVERLAY_DISP++, 0, 0, walletColors[walletUpg].r, walletColors[walletUpg].g, walletColors[walletUpg].b, interfaceCtx->magicAlpha);
#endif
```

include/config/config_game.h
```diff
@@ -1,4 +1,19 @@
...
+#define TEXT_SPEED 2
+
+// Rupee counter symbol changes color with wallet upgrades like in MM
+#define WALLET_COLORS
+
+// Wearing bunnyhood has the same functionality as in MM
+#define MM_BUNNYHOOD
+
+// Configure block push speed multiplier
+#define BLOCK_PUSH_SPEED 3
```
Zelllll, line 14: This should be represented as a float in the define so the user knows that it is one. So, 3.0f, though I would prefer we actually put 1.0f here instead so that the vanilla behavior is the default as the user can choose how fast they want it to be themselves.

include/config/config_game.h
```diff
@@ -1,4 +1,19 @@
 #ifndef CONFIG_GAME_H
 #define CONFIG_GAME_H
 
+// Configure text speed (1 is vanilla)
+#define TEXT_SPEED 2
```
Zelllll, line 5: I'd prefer we call this something like `DIALOGUE_PRINT_SPEED` instead since "text" could be potentially too generic.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -5950,6 +5952,14 @@ void func_8083DDC8(Player* this, PlayState* play) {
 }
 
 void func_8083DF68(Player* this, f32 arg1, s16 arg2) {
+    #ifdef MM_BUNNYHOOD
+
```
Zelllll, line 5956: No need for the new lines between the ifdef and endif here imo

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -5950,6 +5952,14 @@ void func_8083DDC8(Player* this, PlayState* play) {
...
+    if (this->currentMask == PLAYER_MASK_BUNNY) {
+        arg1 *= 1.5f;
```
Zelllll, line 5958: Is it known if this is the exact same behavior as MM?

Zelllll (changes requested):
Just some nitpicks here and there but great job on these changes, these are a great way to start the repo!

src/code/z_map_exp.c
krm01: This actually fixes the pause crash that happens on emulators when pausing with the minimap open!

src/overlays/actors/ovl_player_actor/z_player.c
Recardo-7: MM makes walking 1.5x faster, but I'm not sure if it slows down rolling. Rando seems to do it, but I don`t see the point behind it. But yeah, if it works differently in MM, we should call it `FASTER_BUNNYHOOD` or something instead.

src/code/z_lifemeter.c
Recardo-7: Yes, I personally just find the beeping very distracting and annoying and wanted to replace it with something more subtle, like BOTW does for example. I understand that not everyone wants the blinking though, so I will add a new ifdef for it.

src/code/z_map_exp.c
Recardo-7: Yes, it fixes a bug that causes a crash, so I don't think we need an ifdef here.

src/overlays/actors/ovl_player_actor/z_player.c
Zelllll: Rolling is just the same speed as normal in MM. This is to prevent cheating probably. Only walking forwards is faster with the bunny hood in MM. Any other action is the same speed.

Recardo-7: So should we implement MM behavior or should the player always be faster?

include/config/config_game.h
```diff
@@ -1,4 +1,19 @@
...
+#define WALLET_COLORS
+
+// Wearing bunnyhood has the same functionality as in MM
+#define MM_BUNNYHOOD
+
+// Configure block push speed multiplier
+#define BLOCK_PUSH_SPEED 3
+
+// Replaces the low health beep sound with link blinking red
+#define NO_LOW_HEALTH_BEEP
```
thecozies, line 17: imo negative worded defines can get confusing, especially because you're defining something when it's getting disabled. I'd recommend flipping it like `ENABLE_LOW_HEALTH_BEEP`, and swapping the usage ofc

src/overlays/actors/ovl_player_actor/z_player.c
Zelllll: I’d rather we do whatever MM does for this considering the name of the ifdef. If we can decompile whichever function handles it that would be best so we can recreate it exactly, so this may need a little more research.

Recardo-7: What about the code @Yanis42 shared earlier? Can we use that until the function is decompiled?

include/config/config_game.h
```diff
@@ -4,5 +4,18 @@
 /*****************
  * GAME SETTINGS *
  *****************/
+ 
+// Configure dialogue print speed (1 is vanilla)
+#define DIALOGUE_PRINT_SPEED 2
+
+// Rupee counter symbol changes color with wallet upgrades like in MM
+#define WALLET_COLORS
```
Zelllll, line 12: Maybe we should call this `MM_WALLET_ICON_COLORS` instead.

src/code/z_map_exp.c
```diff
@@ -364,8 +364,8 @@ void Minimap_Draw(PlayState* play) {
 
     OPEN_DISPS(play->state.gfxCtx, "../z_map_exp.c", 626);
 
-    if (play->pauseCtx.state < 4) {
-        switch (play->sceneId) {
+    if (play->pauseCtx.state < 3) {
+        switch (play->sceneNum) {
```
Zelllll, line 368: I think this needs changed here back to `sceneId`, try to merge in the current master branch with your fork.

src/code/z_message_PAL.c
```diff
@@ -3,6 +3,8 @@
 #include "vt.h"
 #include "assets/textures/parameter_static/parameter_static.h"
 
+#include "config/config_game.h"
```
Zelllll, line 6:
```suggestion
#include "config.h"
```

src/code/z_parameter.c
```diff
@@ -4,6 +4,8 @@
 #include "assets/textures/do_action_static/do_action_static.h"
 #include "assets/textures/icon_item_static/icon_item_static.h"
 
+#include "config/config_game.h"
```
Zelllll, line 7:
```suggestion
#include "config.h"
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -8,6 +8,8 @@
 #include "overlays/actors/ovl_Obj_Switch/z_obj_switch.h"
 #include "assets/objects/gameplay_dangeon_keep/gameplay_dangeon_keep.h"
 
+#include "config/config_game.h"
```
Zelllll, line 11:
```suggestion
#include "config.h"
```

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -22,6 +22,8 @@
 #include "assets/objects/gameplay_keep/gameplay_keep.h"
 #include "assets/objects/object_link_child/object_link_child.h"
 
+#include "config/config_game.h"
```
Zelllll, line 25:
```suggestion
#include "config.h"
```

Zelllll (changes requested):
Just a few more suggestions

src/code/z_lifemeter.c
```diff
@@ -508,7 +510,9 @@ void Health_UpdateBeatingHeart(PlayState* play) {
             interfaceCtx->beatingHeartOscillatorDirection = 0;
             if (!Player_InCsMode(play) && (play->pauseCtx.state == 0) && (play->pauseCtx.debugState == 0) &&
                 Health_IsCritical() && !Play_InCsMode(play)) {
```
Yanis42, line 512: is this if really necessary at all? I looked at the functions quickly and it doesn't seem they're doing something else (at least I haven't seen anything obvious)

src/code/z_parameter.c
```diff
@@ -3101,7 +3103,18 @@ void Interface_Draw(PlayState* play) {
         Gfx_SetupDL_39Overlay(play->state.gfxCtx);
 
         // Rupee Icon
+        #ifndef MM_WALLET_ICON_COLORS
```
Yanis42, line 3106: really cool change, how about making a define to let the users choose the color they want? and keep the MM ones by default of course

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -22,6 +22,8 @@
 #include "assets/objects/gameplay_keep/gameplay_keep.h"
 #include "assets/objects/object_link_child/object_link_child.h"
 
+#include "config.h"
```
Yanis42, line 25: is this a left-over? GitHub doesn't show me more changes in this file

Yanis42 (commented):
I really love the changes here, I just have minor thoughts/suggestions here and there

include/config/config_game.h
```diff
@@ -4,5 +4,18 @@
...
+#define DIALOGUE_PRINT_SPEED 2
+
+// Rupee counter icon changes color with wallet upgrades like in MM
+#define MM_WALLET_ICON_COLORS
+
+// Configure block push speed multiplier
+#define BLOCK_PUSH_SPEED 1.0f
+
+// Enables the beep sound at low health
+// #define ENABLE_LOW_HEALTH_BEEP
```
Zelllll, line 18: Should be enabled by default imo

src/code/z_lifemeter.c
```diff
@@ -1,6 +1,8 @@
 #include "global.h"
 #include "assets/textures/parameter_static/parameter_static.h"
```
Zelllll, line 3: I believe we just do config.h now

Zelllll (changes requested):
Just a few more little changes

src/code/z_lifemeter.c
Zelllll: Not sure what you’re saying here. Are you saying the original code should be changed? This looks to be making it so that the low health sound only plays during gameplay.

Recardo-7: I haven't tried to improve the way the original code was written yet (mainly because I would have no idea what I was doing and would probably break everything).
Maybe a big PR that does something like this in multiple places would be nice?

src/overlays/actors/ovl_player_actor/z_player.c
Recardo-7: Yep, when I removed the faster bunnyhood, I forgot this include

Zelllll (approved):
Looks good now!

PR #3, opened 2022-08-05
antialiasing toggle from HackerSM64

ariahiro64:
Everyone knows disabling anti-aliasing is a massive speed boost. Not everyone likes it so i made it act normal by default.


Zelllll (changes requested):
Does HackerSM64 also directly edit `gbi.h`? It would be preferable if we didn't have to but I understand if this is the only way. However one thing that does need changed is that there needs to be a config header added with this option in it. Leave it commented out by default though. I'd suggest making a `config_graphics.h` for this.

Reonu:
> Does HackerSM64 also directly edit `gbi.h`? It would be preferable if we didn't have to but I understand if this is the only way. However one thing that does need changed is that there needs to be a config header added with this option in it. Leave it commented out by default though. I'd suggest making a `config_graphics.h` for this.

Yes, it does. It's pretty much the only way to make sure that AA is disabled on every single material, even if you don't explicitly disable it on fast64 or whatever. It has been tested to death and it doesn't cause any issues.

Zelllll:
Alright then in that case no problem. We just need `config_graphics.h` and then this will be good to go.

include/config/config_graphics.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
Zelllll, line 1: For this project I’d rather we use traditional include guards

Zelllll (changes requested):
Just change how you do the include guards and this is good 👍

include/config/config_game.h
```diff
@@ -1,4 +1,6 @@
 #ifndef CONFIG_GAME_H
 #define CONFIG_GAME_H
 
+#include "config_graphics.h"
```
thecozies, line 4: is there a base config file? I would expect include/config.h to include all of these sub config files

thecozies (commented):
part of the gbi stuff was missed here and this may catch more AA cases than the other chunk

https://github.com/HackerN64/HackerSM64/blob/a549e1a95a50f7b48d9cd655d8b82ba05f0d2e6d/include/n64/PR/gbi.h#L735

AriaHiro64:
last commit cause tranparency issues so im double checking my work. dont merge yet

AriaHiro64:
ok just checked and IM_RD is required for oot.

Arceveti:
Why is the `|` part of the define? Why not just define it to 0 if `DISABLE_AA` is enabled?

AriaHiro64:
wasnt done experimenting. should be ready except for the changes zel suggested

AriaHiro64:
> Why is the `|` part of the define? Why not just define it to 0 if `DISABLE_AA` is enabled?

thats how its setup in hackersm64. also some transparency effects wont work with that method as aa is needed.

Arceveti:
> > Why is the `|` part of the define? Why not just define it to 0 if `DISABLE_AA` is enabled?
> 
> thats how its setup in hackersm64. also some transparency effects wont work with that method as aa is needed.

It should be fixed in HackerSM64 too. It shouldn't be possible for this to affect any sort of trasnparency effects, since it's just formatting and compiles to the exact same thing. Unless you try to use `AA_DEF`/`RD_DEF` on their own, in which case the current method will just give a syntax error. The current method also means that every render mode define that uses it is missing a `|`, which slightly hinders readability. Removing the `|` from the `AA_DEF`/`RD_DEF` defines and restoring them in the render mode defines, and defining `AA_DEF`/`RD_DEF` to 0 instead of blank when `DISABLE_AA` is enabled, should fix both of those issues.

AriaHiro64:
ive already tried it and transparency becomes a color. usually black

AriaHiro64:
oh wait i read that wrong

AriaHiro64:
> > > Why is the `|` part of the define? Why not just define it to 0 if `DISABLE_AA` is enabled?
> > 
> > 
> > thats how its setup in hackersm64. also some transparency effects wont work with that method as aa is needed.
> 
> It should be fixed in HackerSM64 too. It shouldn't be possible for this to affect any sort of trasnparency effects, since it's just formatting and compiles to the exact same thing. Unless you try to use `AA_DEF`/`RD_DEF` on their own, in which case the current method will just give a syntax error. The current method also means that every render mode define that uses it is missing a `|`, which slightly hinders readability. Removing the `|` from the `AA_DEF`/`RD_DEF` defines and restoring them in the render mode defines, and defining `AA_DEF`/`RD_DEF` to 0 instead of blank when `DISABLE_AA` is enabled, should fix both of those issues.

How's that?

AriaHiro64:
accidetly ckicked cose

AriaHiro64:
How's That? Everything should be contained in gbi.h and the main config file.

Yanis42 (approved):
(I don't understand a lot of things about graphics)

Yanis42:
This branch has conflicts, when those are fixed I'll try this branch and merge it if there's no issues (though I'd be nice if someone who knows a bit about graphics review this PR)

Yanis42:
I read the conversation a bit and it seems like the ``config_graphics.h`` was deleted, I think you should add it back and add the safeguards (like ``#ifndef CONFIG_GRAPHICS_H`` etc), and the include in ``include/config.h``

otherwise I don't have much to say on this PR as I don't understand DL commands

PR #4, opened 2022-08-06
Pause delay fix that works on console

ariahiro64:
Have no idea why this code is called but it does nothing but lag emulators. Removing it is a slight speed boost on console and a major on emulator.


Yanis42:
PR #2 has a fix for the pause menu I believe

Recardo-7:
> PR #2 has a fix for the pause menu I believe

It fixed the crash that sometimes occurs when pausing. Didn't speed it up.

src/code/PreRender.c
```diff
@@ -551,9 +431,6 @@ void PreRender_Calc(PreRender* this) {
 
                 a >>= 5;
                 a++;
-                if (a != 8) {
```
Zelllll, line 554: If you look closely this isn't really the best way to do this. You removed the part of the loop that is actually doing something, and the rest of this function is now completely useless. You can just never call `PreRender_Calc` and remove this function and any other functions that it calls.

Zelllll (changes requested):

AriaHiro64:
closing this for now because someone else has already made a pull request with it and i didnt realize it

Recardo-7:
> 

#2 just fixed a crash that happens when pausing with the minimap open. It **did not** remove this function to speed up pausing.

AriaHiro64:
actally ill keep it open just incase im doing somthing different

AriaHiro64:
Rebased and did suggested changes so review when you can.

Yanis42:
I'm worried about something I need to try myself about this PR, I'll review it once I tried

Yanis42 (approved):
Works on N64, my concern was about frame advancing the game using the pause menu, which can be really important sometimes if you wanna be the very best (like no one ever was)

src/code/z_play.c
```diff
@@ -1112,7 +1112,6 @@ void Play_Draw(PlayState* this) {
 
             if (R_PAUSE_MENU_MODE == 2) {
                 Sched_FlushTaskQueue();
```
Zelllll, line 1114: I believe you can remove this line too.

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -256,12 +256,6 @@ void KaleidoScope_SetupPlayerPreRender(PlayState* play) {
     CLOSE_DISPS(play->state.gfxCtx, "../z_kaleido_scope_PAL.c", 509);
 }
 
-void KaleidoScope_ProcessPlayerPreRender(void) {
```
Zelllll, line 259: Removing this here isn't entirely correct since it also calls the destroy function.

PR #5, opened 2022-08-06
allocate more memory to gi models since the amount was very tiny

ariahiro64:
you can do huge ones like this now
image


Recardo-7:
Maybe you could make all your changes optional and put them in the config file, because you never know what some people want in their hacks or what conflicts with the code they wrote. I'd rather have tons of options in the config file than not be able to disable something I might not want.

AriaHiro64:
I possibly could have people insert their desired amount but have examples of save numbers and keep the vanilla values as the default.

Zelllll:
Yeah, this should be something inside a config file. maybe a `config_system.h` that handles allocation-related things?

AriaHiro64:
That should be acceptable for now. Tell me if there are further changes needed. It's normal by default in case people don't need to increase it

include/config/config_game.h
```diff
@@ -4,5 +4,17 @@
...
+/*****************
+*** Allocation ***
+/****************/
+/*******************************************
+*GI models or "Get Item" models are*********
+*the models that appear over Link's head****
+*when he picks up a item. the normal memory*
+*allocated for the is 0x3008. a safe number*
+*is 1024 * 32 which is way more than normal*
+*******************************************/
```
Yanis42, line 17:
```suggestion
/*************************************************************
 * GI ("Get Item") models, are those that appear             *
 * over Link's head when he picks up an item.                *
 * The normal memory allocated for the is 0x3008.            *
 * A safe number is 1024 * 32, which is way more than normal.*
 *************************************************************/
```
Looks better like that imo, also I think it'd be nice to precise the unit, I assume it's bytes but I wasn't sure so I didn't mention it in the suggestion

AriaHiro64: looks better. i also used 1024 * 32 because an exact value may seem like a weird arbitrary value to newcomers.

AriaHiro64: like 32768 as opposed to 1024 * 32

AriaHiro64: i also think 32 kb might be overkill but its like the biggest value i tested and seems safe. bigger hack may need smaller sizes.

Yanis42: I still think we should add the unit in the comment

AriaHiro64: i could but zel said he may do this dynamically so ima wait for him

Zelllll: Can you test #18 to see if it works for you with your larger GI models? @AriaHiro64

ariahiro64:
#18 is a better implementation.

PR #6, opened 2022-08-07
Add build date

Zelllll:
Adds the build date using `__TIME__` and adds build options, This is normally displayed in-game through the N64 logo screen.

### DIFF
--- a/src/boot/build.c
+++ b/src/boot/build.c
@@ -1,3 +1,7 @@
-const char gBuildTeam[] = "zelda@srd022j";
-const char gBuildDate[] = "03-02-21 00:16:31";
+const char gBuildTeam[] = "HackerOOT";
+const char gBuildDate[] = __TIME__;
+#ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


Reonu:
Ehhhhhh be careful with this one. Make it debug only or something. It prevents reproducible builds which is bad

Zelllll:
Maybe we can add a deterministic build flag or something.

Reonu:
> Maybe we can add a deterministic build flag or something.

I think the only time when it should *not* be deterministic is when you're making a debug build. So maybe just ifdef it with a debug build flag. Just my two cents though, it's your repo! I'm just saying it because there have been problems with HackerSM64 builds not being deterministic in the past.

Zelllll:
Yeah, I certainly understand the concern, though I'd like if all builds did have a date attached, even if not debug.

Reonu:
testing that the github discord channel works

Zelllll:
Closing this for now but will revisit later.

PR #8, opened 2022-08-07
Restore N64 boot logo, add DEBUG_ROM define to Makefile, and more

Zelllll:
- Restores N64 boot logo which was removed in GCN builds (which decomp is based on).
- Prints build info on the N64 logo...
- The ROM now stores the build date, author name, and commit hash.
- Add DEBUG_ROM define, which can be set in the Makefile
- Change how the configs are handled...
- There is now a single `config.h` in the include folder which replaces `config_all.h` and should be included instead of individual config headers now.
- There is now a `config_debug.h` which includes @kurethedead 's scene bootup features and is now only enabled if `DEBUG_ROM` is defined.
- There is now a safeguard header that handles conflicting config features.
- Change region so the game will run at 20fps instead of ~15fps.


Zelllll:
test

Reonu (approved):
it's a bit weird to me that non-debug builds are not deterministic by default but eh

Zelllll:
imo they don't need to be since being deterministic is only important in the context of debugging

PR #9, opened 2022-08-07
Add options to boot directly to a scene/file select

kurethedead:


kurethedead:
Actually, hold off on this PR for a moment, trying to figure out some issues.

kurethedead:
All right, issues should be ironed out now.

Zelllll (approved):
For now I’m fine with this but eventually I want to refactor z_sram so this may change a little.

PR #11, opened 2022-08-22
Fix: signed byte bug for scene id in entranceinfo

krm01:
having this be a signed byte means you're capped at 128 scenes. OG devs should have made this unsigned.


Zelllll (approved):
Looks good!

PR #13, opened 2022-08-23
Feature: "Split Farore's Wind" Option

Yanis42:
Inspired by OoT Randomizer, adds a define to have one FW point as child and one as adult, vanilla OoT resets the point when you're traveling in time

Let me know if this is a good idea or not, I'm not sure to fully understand what features we wanna add for HackerOoT, I just thought modders may want this one

Any feedback is welcome!


include/z64save.h
```diff
@@ -132,7 +134,7 @@ typedef struct {
     /* 0x0068 */ ItemEquips equips;
     /* 0x0074 */ Inventory inventory;
     /* 0x00D4 */ SavedSceneFlags sceneFlags[124];
-    /* 0x0E64 */ FaroresWindData fw;
+    /* 0x0E64 */ FaroresWindData fw_main;
```
Zelllll, line 137: should be camelCase, so fwMain. Same goes for fwSecondary. Could also just make this an array if you want.

include/z64save.h
```diff
@@ -212,7 +214,10 @@ typedef struct {
     /* 0x1420 */ s16 worldMapArea;
     /* 0x1422 */ s16 sunsSongState; // controls the effects of suns song
     /* 0x1424 */ s16 healthAccumulator;
-} SaveContext; // size = 0x1428
+#ifdef FW_SPLIT_AGE
+    /* 0x1428 */ FaroresWindData fw_secondary;
```
Zelllll, line 218: I'd rather this be next to the original variable. I know this will change the structure of SaveContext but that is going to have to happen regardless, so if that breaks any speedrunning glitches or something it is a necessary evil for this project, imo.

src/code/z_demo.c
```diff
@@ -488,6 +490,10 @@ void Cutscene_Command_Terminator(PlayState* play, CutsceneContext* csCtx, CsCmdB
     Player* player = GET_PLAYER(play);
     s32 temp = 0;
 
+#ifdef FW_SPLIT_AGE
+    FaroresWindData fw_backup;
```
Zelllll, line 494: camelCase here again

src/code/z_demo.c
```diff
@@ -558,10 +564,15 @@ void Cutscene_Command_Terminator(PlayState* play, CutsceneContext* csCtx, CsCmdB
...
+#ifdef FW_SPLIT_AGE
+                fw_backup = gSaveContext.fw_main;
+                gSaveContext.fw_main = gSaveContext.fw_secondary;
+                gSaveContext.fw_secondary = fw_backup;
+#else
+                gSaveContext.fw_main.set = 0;
+#endif
                 gSaveContext.respawn[RESPAWN_MODE_TOP].data = 0;
                 if (!GET_EVENTCHKINF(EVENTCHKINF_45)) {
-                    SET_EVENTCHKINF(EVENTCHKINF_45);
```
Zelllll, line 564: How come this gets removed, and is removed in all code paths, regardless of the ifdef?

Zelllll (changes requested):
Mainly syntax things but I have a few other suggestions.

include/z64save.h
Yanis42: tbh I was just worried about shifting savecontext raising issues, I believe it's still a bit weird how it's handled in decomp, but yea it would be better next to the original variable

Zelllll: If there is an alternate struct in z_sram.c for this still, then just update that one too for now. I'm going to have to go through and fix the save context issues soon.

Yanis42: (on a side note: I don't think there's any glitch that requires the savecontext, except SRM but I expect this to be patched on this repo as it can be problematic in mods)

src/code/z_demo.c
Yanis42: ah I think it's my bad on this one, during my tests I set some flags and I guess I thought this one among the flags I added, gonna revert this

Zelllll (approved):

PR #14, opened 2022-08-24
[Feature] Map Select Changes

Yanis42:
I decided to change stuff on the map select to make it more modder-friendly:

List of changes:
- Remove "OPT" as it's unused
- "Stage" -> "Scene Layer" ("scene layer" is what you may call "scene setup")
- Translated everything to english (thanks to Louis for the scene names)
- Named some generic names (such as "Grotto 1", "Grotto 2", things like that)
- Changed the color of "Link's Age" and "Scene Layer" (I'm not sure this is a good change but it's less yellow at least :kappa~2:)
- Moved a bit the position of the scene selection part to the left
- "ZELDA MAP SELECT" -> "Zelda Map Select" (couldn't think of a better name)
- Removed the numbers from the scene name strings, instead it's added with the scene number mathed in the printing loop inside MapSelect_PrintMenu
- Vertical alignment of the numbers to improve readability
- Inverted "Link's Age" and "Scene Layer" positions
- Accessibility changes:
-> Added an arrow next to the selected scene name
-> C-Left and C-Right change the color of the selected scene name
- Fixed sounds not playing (the issue was the volume turned down by the file select)
- Add sounds for (controls) page, Link's age (random) and scene layer changes
- Added a BGM (Owl Theme as Adult BGM), it can be turned off in config_debug.h
- Added an option to boot directly on map select
- Made the file more readable

<details closed>
<summary>Main Page</summary>

</details>
<details closed>
<summary>Controls Page</summary>

</details>


Yanis42:
There's some issues with the map select currently:
- grotto crash when you exit (make sense since you didn't enter it from another scene)
- the game doesn't check if the cutscene number is correct, so again you can get a crash
- you can't choose the spawn currently (not really important but I'll add that later)

I'm going to fix those Soon™

src/overlays/gamestates/ovl_select/z_select.h
```diff
@@ -0,0 +1,258 @@
+#ifndef Z_SELECT_H
```
Zelllll, line 1: Shouldn't this file also include the struct for MapSelectState?

Zelllll (changes requested):
Forgot to submit this review! Overall this is good but there may be a point where we completely redo the map select entirely. In my hack that I am working on my map select is dynamic where it will update itself whenever new scenes are added to the game. At some point we should do that here I think.

src/overlays/gamestates/ovl_select/z_select.h
Yanis42: iirc the headers make this impossible, I can't remember what's the real reason though, but I remember trying that and having issues

Zelllll: It shouldn't be "impossible" but with how weird all the includes are right now in the repo we can focus on this later instead.

Yanis42: I think the only issue is getting the gamestate struct without having looping includes, otherwise it should be fine

Zelllll (approved):
Looks good to me! Nice job!

PR #15, opened 2022-08-28
[Feature] Defines for debug rom's features

Yanis42:
This PR adds defines to enable or disable a lot of features from the debug rom. There's currently 15 features but there might be more than that, here's the list of which feature I worked on:

## General Features
- No Clip
- Input Display
- Cutscene control (skip, restart, stop, infos print)
- Time of day informations
- Framerate options (frame advance, pausing)
- Map Select (see PR #14 for the upgraded version)
- Load debug save on file 1
- Remove test scenes

## In-Game Editors
- Inventory Editor
- Event Editor (for flags, this might need a rename)
- Registers Editor (this is the heart of the debugging system, it was used to enable some feature in-game, it might hide some features I don't know about)

## In-Game Debuggers
- Camera Debugger (includes a cutscene creator which has an option to export to a mempak)
- Audio Debugger (I might have missed useless code to ifdef, this one is really tough to understand)
- Actor Debugger (shows arrows and other things related to actor, hence the size of this PR)
- Message Debugger (shows the text ID when a message is on-screen, you can also show any dialog you want)

All of the above can be disabled with the ``DISABLE_DEBUG_FEATURES`` define, which gets enabled automatically when it's in release mode.

I used ifdefs to remove every piece of code related to each feature, including removing from the spec file, hence not include it in the compiled rom, it just makes sense to not include useless stuff to me. Also I think we can remove ``z_select_static`` as it seems completely unused but I'm not 100% sure.

By doing that I encountered one issue with the entrance table, ZAPD doesn't use the enum names for the exit lists so I had to make a temp fix to avoid shifting the table.

In another PR I'll enhance a bit some of those features, though this PR has some minor enhancements, like allow control stick inputs for moving in no clip mode, in another PR I'll do more.

Also I made custom control mappings for features that need to use a controller, you can also choose which controller you want to use.

I think that's everything I have to say about this for now, as always if you have any feedback feel free!

PS: sorry for the big PR, but it's not *that* big considering half of the changed files are overlays with 2-3 edits, also this time I thought about not doing a big commit for everything, so you can look at the commit history

Special thanks:
- [TCRF](https://tcrf.net), which helped me a lot for understanding and testing the features I worked on
- [Fig](https://github.com/fig02) for the z_demo documentation on the decomp repo, which helped me figuring out stuff in the cutscene system


include/config/config_debug.h
```diff
@@ -34,4 +37,45 @@
...
 #define BOOT_PLAYER_NAME 0x15, 0x12, 0x17, 0x14, 0x3E, 0x3E, 0x3E, 0x3E // "LINK"
 
+/**
+ * Disable every debug features
+ */
+// #define DISABLE_DEBUG_FEATURES
+
+// -------------------------------------------
+
+// TODO: find a better way to do that
```
Yanis42, line 47: About that: initially I put the undefs in the safeguard file, but for the global toggle define I also needed to undef everything, so to avoid duplicates I basically moved the safeguards for everything I added in the config_debug file, also it was easier to enable the "file 1 normal" thing

Yanis42:
I knew I was forgetting something, I made a ``ENABLE_FAST_TEXT`` define, but it's a bit out of the scope of this PR

include/config/config_debug.h
```diff
@@ -34,4 +37,45 @@
...
+#ifndef DEBUG_ROM
+#define DISABLE_DEBUG_FEATURES
+#endif
+
+#ifdef DISABLE_DEBUG_FEATURES
+#undef SHOW_CS_INFOS
+#undef SHOW_INPUT_DISPLAY
+#undef SHOW_TIME_INFOS
+#undef NO_TEST_SCENES
```
Zelllll, line 56: Wouldn't this be a "retail" feature? Since the test scenes are a debug feature?

include/config/config_debug_controls.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Change the debug features' controls
+ */
+
+// -------------------------------------------
+
+/**** [NO CLIP CONTROLS] ****/
+// Controller port to use
+#define NOCLIP_CONTROLLER_PORT CONTROLLER_PORT_1
```
Zelllll, line 16: This is a great idea here! I actually hope that at some point we can do the same thing for z_player as a whole, so we can easily remap the buttons for different actions.

include/config/config_debug_toggles.h
```diff
@@ -0,0 +1,82 @@
...
+
+/**** [TIME INFORMATIONS] ****/
+// #define SHOW_TIME_INFOS
+
+/**** [FRAMERATE OPTIONS] ****/
+// TODO: check for frame advance for drawing, improve how it works
+#define ENABLE_FRAMERATE_OPTIONS
+
+/**** [MAP SELECT] ****/
+// #define ENABLE_MAP_SELECT
```
Zelllll, line 40: For this, we should probably ifdef the entire map select gamestate, meaning that it should be ifdef'd out in the spec as well, and in the gamestate table.

include/z64cutscene.h
```diff
@@ -97,11 +97,11 @@ typedef struct {
 } CsCmdActorAction; // size = 0x30
 
 typedef enum {
-    CS_STATE_IDLE,
-    CS_STATE_SKIPPABLE_INIT,
-    CS_STATE_SKIPPABLE_EXEC,
-    CS_STATE_UNSKIPPABLE_INIT,
-    CS_STATE_UNSKIPPABLE_EXEC
+    CS_STATE_IDLE,              // CS_STATE_IDLE
```
Zelllll, line 100: Can you explain the changes here?

spec
```diff
@@ -8439,6 +8453,7 @@ beginseg
     number 3
 endseg
 
+#ifndef NO_TEST_SCENES
```
Zelllll, line 8456: Might be better to just rearrange the spec to have these all together.

src/code/main.c
```diff
@@ -56,6 +59,8 @@ void Main(void* arg) {
     // "System heap initalization"
     osSyncPrintf("システムヒープ初期化 %08x-%08x %08x\n", systemHeapStart, fb, gSystemHeapSize);
     SystemHeap_Init((void*)systemHeapStart, gSystemHeapSize); // initializes the system heap
+
+#ifdef ENABLE_CAMERA_DEBUGGER
```
Zelllll, line 63: You need to be a little careful here. Although the debug heap is only used for the camera debugger, it can technically be used for other things, so maybe it's best that we leave this code alone for the time being.

src/code/mempak.c
```diff
@@ -1,3 +1,7 @@
+#include "config.h"
+
+#ifdef ENABLE_CAMERA_DEBUGGER
```
Zelllll, line 3: Can't the mempak technically be used for other things? Might be better not to remove these entirely just because it's not used by the vanilla game. Also it would be better to ifdef this in the spec instead of in the body of the file, imo.

src/code/z_debug_display.c
```diff
@@ -1,3 +1,7 @@
+#include "config.h"
+
+#ifndef NO_DEBUG_DISPLAY
```
Zelllll, line 3: Same thing here, may be better to just ifdef this entire file in the spec rather than in the body of the file.

src/code/z_demo.c
```diff
@@ -118,12 +120,13 @@ void Cutscene_DrawDebugInfo(PlayState* play, Gfx** dlist, CutsceneContext* csCtx
     GfxPrint_SetColor(&printer, 255, 255, 255, 32);
     GfxPrint_Printf(&printer, "%06d", csCtx->frames);
     GfxPrint_SetColor(&printer, 50, 255, 255, 60);
-    GfxPrint_SetPos(&printer, 4, 26);
+    GfxPrint_SetPos(&printer, 10, 26);
```
Zelllll, line 123: Was this an intentional change here?

src/code/z_effect_soft_sprite.c
```diff
@@ -155,12 +156,16 @@ s32 EffectSs_FindSlot(s32 priority, s32* pIndex) {
 void EffectSs_Insert(PlayState* play, EffectSs* effectSs) {
     s32 index;
 
+#ifdef ENABLE_FRAMERATE_OPTIONS
     if (FrameAdvance_IsEnabled(play) != true) {
```
Zelllll, line 160: We could probably just create a macro for this check here. For example
```
if (FRAME_ADVANCE_DISABLED()) {
    ...
}
```
In the debug code path this would of course be defined as the same check here, but otherwise would just simply be defined as true.
This also goes for some other things in this PR I think, so it is worth considering creating other macros like this for other debug specific checks.

Zelllll: Now that I am thinking about this more, we could just simply make a generic `DBG_COND()` macro, which takes the condition as it's input, so then we could use it everywhere.

Zelllll (changes requested):
Awesome job on this, I really like the approach you took with defines for each debug feature. I've left some comments to start off with, but I will give this a second pass after you resolve them.

src/code/mempak.c
Yanis42: afaik the mempak is only used for the cutscene exporter, when I did this I only thought about removing each debug feature and wasn't thinking about another use case, I'll change that and make its own define for the mempak it seems the best option 👍 

(so yea technically it could be used for something else, though idk how you'd use it, I don't even know how are you supposed to get a mempak's data)

src/code/z_debug_display.c
Yanis42: does ifdeffing in spec "prevent" the compilation of the file? because that's the only reason why I did this to begin with, compiling code you won't use isn't great

src/code/z_demo.c
Yanis42: I think it's a leftover from an attempt to enhance a bit those features, I made a separate branch for those kind of changed tho so I can revert this, I can't remember why I changed that tho

include/config/config_debug.h
```diff
@@ -34,4 +37,45 @@
...
+#undef ENABLE_CAMERA_DEBUGGER
+#undef ENABLE_AUDIO_DEBUGGER
+#undef ENABLE_ACTOR_DEBUGGER
+#undef ENABLE_MSG_DEBUGGER
+#undef ENABLE_DEBUG_SAVE
+#endif
+
+// Remove map select from file 1
+#if (defined BOOT_TO_SCENE && defined BOOT_TO_FILE_SELECT) || (!defined ENABLE_MAP_SELECT) || (defined DISABLE_DEBUG_FEATURES)
+#define FILE_1_NORMAL
```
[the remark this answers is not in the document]
Yanis42: finding names isn't something I'm good at as you can see lol, ``DEBUG_FILE_1`` sounds better 👍

src/code/z_debug_display.c
Zelllll: That’s fair I guess. It would prevent linking it. I just don’t like the idea of linking empty files a whole lot is all but perhaps you’re right. We can keep it how you have it here.

include/config/config_debug.h
Yanis42: you mean move ``NO_TEST_SCENES`` in config_game.h? I'm not sure to understand what you mean exactly

include/config/config_debug_controls.h
Yanis42: yea I wanted to give modders an easy way to change every controls and controller port, there's documentation on TCRF (which helped me a lot) but it's better if you can customise everything imo

also I try my best at making OoT accessible for everyone 😄

src/code/z_debug_display.c
Yanis42: tbh I just forgot the spec file existed at all lmao, it's weird to link empty files I agree, I'll add more ifdefs in spec and re-arrange the order like you mentioned, creating a "debug" segment in spec might be worth

Zelllll: Well the spec has to be ordered a certain way for some things, but if you want you can reorder where the .o's for debug related things are linked in `code`, if that makes sense.

include/z64cutscene.h
Yanis42: I used the documentation Fig did for decomp to help me figure out how z_demo was working, I just forgot to remove the extra stuff, I'll revert this considering this documentation is an opened PR on decomp so it'll be merged here someday

include/config/config_debug.h
Zelllll: Why are we *un*defining NO_TEST_SCENES when debug features are off? That would mean that we are including test scenes for retail builds, and not debug builds, it seems.

Zelllll: The solution here btw should be to make an `INCLUDE_TEST_SCENES` instead of this. (so then it's not a negative define)

src/code/main.c
Yanis42: 👍 I wasn't sure if removing the debug heap was a good idea anyway

include/config/config_debug.h
Yanis42: oh yea, it's clearly a mistake then, debug scene shouldn't be part of retail

include/config/config_debug_toggles.h
Yanis42: I already did that, I added ``GAMESTATE_NONE`` as a temp workaround since the table has hardcoded indices

src/code/z_demo.c
Yanis42: oh yea I remember now, I changed that because I couldn't see the ``"SKIP=(START) or (Cursole Right)"`` text on screen, so I thought maybe it was behind the black bars, reverted 👍

include/config/config_debug_toggles.h
Zelllll: Yup ended up realizing that later on in my review and forgot to remove this comment. As an aside we ought to completely redo that table. I did so in my current hack I'm working on and made it work in a much more sane way lol

Yanis42:
more conflicts 😔, I'll fix it later and if everything works properly I'll merge this

once conflicts are gone it'd be nice of someone tries this branch to see if something's wrong 👍

Yanis42:
conflict fixed, while doing this I noticed the reg editor stuff (and probably audio too) might not be 100% accurate, otherwise it built on my end

ariahiro64 (approved):

Yanis42:
merging this tomorrow 👀

PR #17, opened 2022-09-04
[Feature] Introduce Assets Fixer

Yanis42:
This is a small program to fix assets, this will:
- add casts to room's symbols
- remove hexadecimal from exit lists, instead it's reading the entrance table
- use new names, ``CamData`` becomes ``BgCamInfo``, etc

I made this alone so I'm open to any feedback

Note that this is a temp fix until ZAPD is updated

Everything about it is detailed [there](https://github.com/Yanis42/DecompAssetsFixer#decompilation-assets-fixer-daf)

---
Tested under Python 3.10, rom built OK with decomp


Zelllll (approved):
This looks good to me. I didn't read all the python but since this tool is pretty modular from the rest of the repo I think this is a fine addition.

Yanis42:
forgot to fix a compatibility issue with older Python versions, this should now be fixed (the issue was the method I used to merge dictionnaries, since Python 3.9 you can do ``dict1 | dict2``, so older versions need [something else](https://peps.python.org/pep-0584/#dict-d1-d2))

Zelllll:
Nice job. I'm going to go ahead and test this on my machine later this evening and if it works fine then I'll go ahead and merge this.

Yanis42:
I need to fix the conflicts and try this again, I'd be nice if someone else tries this and confirms it works as expected

Yanis42:
conflicts are fixed, this needs someone else to try this before merging

Yanis42:
I setup a WSL installation with Debian and Python 3.7(.15) and it worked properly, gonna merge this tomorrow if nothing comes up

Also I confirmed the entrance table can be shifted, it works fine

ariahiro64 (approved):
seems good enough

PR #18, opened 2022-09-11
Dynamically Allocate the Get Item Segment

Zelllll:
This PR adds code in z_player.c to dynamically choose the size of the getItemSegment based on whatever the largest get item object is.

### DIFF
--- a/include/z64player.h
+++ b/include/z64player.h
@@ -383,6 +383,8 @@ typedef struct {
     /* 0x10 */ Vec3f base;
 } WeaponInfo; // size = 0x1C
 
+#define PLAYER_GI_SEG_MIN 0x2880 // max size of a place title card file
+
 #define PLAYER_STATE1_0 (1 << 0)
 #define PLAYER_STATE1_SWINGING_BOTTLE (1 << 1)
 #define PLAYER_STATE1_2 (1 << 2)
@@ -491,6 +493,7 @@ typedef struct Player {
     /* 0x0194 */ OSMesgQueue giObjectLoadQueue;
     /* 0x01AC */ OSMesg     giObjectLoadMsg;
     /* 0x01B0 */ void*      giObjectSegment; // also used for title card textures
+                 u32        giObjectSegmentMaxSz;
     /* 0x01B4 */ SkelAnime  skelAnime;
     /* 0x01F8 */ Vec3s      jointTable[PLAYER_LIMB_BUF_COUNT];
     /* 0x0288 */ Vec3s      morphTable[PLAYER_LIMB_BUF_COUNT];
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -9288,6 +9288,20 @@ static void (*D_80854738[])(PlayState* play, Player* this) = {
 
 static Vec3f D_80854778 = { 0.0f, 50.0f, 0.0f };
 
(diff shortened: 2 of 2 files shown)


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9288,6 +9288,20 @@ static void (*D_80854738[])(PlayState* play, Player* this) = {
 
 static Vec3f D_80854778 = { 0.0f, 50.0f, 0.0f };
 
+u32 Player_GetLargestGI(void) {
+    u32 maxObjSize = 0, curObjSize = 0;
+    for (s32 i = 0; true; i++) {
```
Yanis42, line 9293: I'm not a huge fan of infinite loops tbh, well it's not an infinite loop since there's a return but idk, I'd prefer something like this:
```suggestion
    for (s32 i = OBJECT_ID_MAX; i == OBJECT_INVALID; i--) {
```
the idea is starting from the the max ID and going in reverse through the object list, if I'm understanding this correctly the loop goes through the objects to get the max size value

include/z64player.h
```diff
@@ -383,6 +383,8 @@ typedef struct {
     /* 0x10 */ Vec3f base;
 } WeaponInfo; // size = 0x1C
 
+#define PLAYER_GI_SEG_MIN 0x2880 // max size of a place title card file
```
Yanis42, line 386: maybe this could go in config_game.h? unless the value should always be 0x2880

Yanis42 (commented):
I don't understand a lot when it comes to memory stuff but I do have some suggestions for the code itself

src/overlays/actors/ovl_player_actor/z_player.c
Yanis42: well, it should be ``s32 i = OBJECT_ID_MAX - 1`` actually but you get the idea

AriaHiro64:
image

include/z64player.h
AriaHiro64: i dont see people touching this value if the game auto updates it

Yanis42: if users shouldn't edit this then this is fine yeah

src/overlays/actors/ovl_player_actor/z_player.c
Zelllll: I think you’re misunderstanding. The loop iterator is not the object ID, that is being read from the table, so this code would not work as is.

Yanis42: oh yea you're right lol, so there's no alternative to this pseudo infinite loop?

Zelllll: Not really, you could loop as many times as there are entries in the table, but the table ends when there is OBJECT_ID_INVALID, which is reflected in the code here. I personally feel that this is the most elegant way to do it, hence why I did it this way, but open to suggestions.

include/z64player.h
Zelllll: This is more of a placeholder for now until I can think of a better way to handle it. The get item buffer is also used for the place name textures, which ironically are actually larger than any of the vanilla get item objects.

Yanis42 (approved):
Approving since both discussions I opened got resolved 👍

Zelllll:
Something that I just noticed is that the new variable added to the player struct is not really necessary, so I’ll remove that before merging this.

Issue #19, opened 2022-11-23
Add compiler flag needed for project to compile


sauraen:
`mips-linux-gnu-gcc (Debian 10.2.1-6) 10.2.1 20210110`

`mips-linux-gnu-gcc` will not compile any file with the `-mno-abicalls` option unless the `-fno-pic` option is also specified. For example:

```
>~$ touch test.c
>~$ mips-linux-gnu-gcc -mno-abicalls test.c -o test.o
cc1: error: position-independent code requires ‘-mabicalls’
```

This causes every file being compiled in this project (for N64) to fail. The solution is to add `-fno-pic`; with this change, the project compiles and works on console.

```
diff --git a/Makefile b/Makefile
index d62f7179b..748d495cf 100644
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ endif
 ASFLAGS := -march=vr4300 -32 -no-pad-sections -Iinclude
 
 ifeq ($(COMPILER),gcc)
-  CFLAGS += -G 0 -nostdinc $(INC) -march=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks -fno-zero-initialized-in-bss -fno-toplevel-reorder -ffreestanding -fno-common -fno-merge-constants -mno-explicit-relocs -mno-split-addresses $(CHECK_WARNINGS) -funsigned-char
+  CFLAGS += -G 0 -nostdinc $(INC) -march=vr4300 -mfix4300 -mabi=32 -fno-pic -mno-abicalls -mdivide-breaks -fno-zero-initialized-in-bss -fno-toplevel-reorder -ffreestanding -fno-common -fno-merge-constants -mno-explicit-relocs -mno-split-addresses $(CHECK_WARNINGS) -funsigned-char
   MIPS_VERSION := -mips3
 else 
   # we support Microsoft extensions such as anonymous structs, which the compiler does support but warns for their usage. Surpress the warnings with -woff.
```

I'm not sure why no other developer has encountered this error yet. However, adding `-fno-pic` should not be harmful if someone has a version of `mips-linux-gnu-gcc` which does not default to position independent code enabled.

Dragorn421:
fwiw this has been encountered before
Link to in the oot decomp server https://discord.com/channels/688807550715560050/688851337085190255/949331416607514634

> A friend of mine compiled with gcc on an AArch64 Debian system and it throws the following error:
> ```
[...]

PR #24, opened 2022-12-04
Add required compiler flag

Yanis42:
Fix for #19, I never had such issues on my end (and I guess no one else on main decomp too since it's not there), tried it (not on console) and it seems fine so it should be fine 👍


Yanis42:
merging this tomorrow

PR #25, opened 2022-12-04
[Feature] Dynamic Get Item object size allocation

Yanis42:
Basically a replacement for #18 (with different code of course), seems to work properly on my end


Yanis42:
idk if doing ``size + 16`` is really necessary but I was thinking without it the align could eat a bit the required size, so I guess it's good to have for safety

ariahiro64 (changes requested):
```ASSERT(size <= 1024 * 8, "size <= 1024 * 8", "../z_player.c", 9091);```
this can be completely removed since this just crashes after the original limit is hit

Yanis42:
should be good now 👍 

thanks!

ariahiro64 (approved):

Yanis42:
merging soon™ (tomorrow)

PR #28, opened 2022-12-08
fix typecasting for binary angles

ariahiro64:


Yanis42:
there's a parenthesis issue

PR #35, opened 2022-12-13
Ifdef more debug things

Yanis42:
Toggles for speedmeter and the debug heap, note that the debug heap will turn on if the camera debugger is enabled as it's needed for it

now that I think about it it might be worth to turn the "debug heap" into something you can use for your mod but idk it's ok for now imo

also I want to ifdef unused assets to save extra space in the rom


ariahiro64 (approved):
tried this out for an hour and didnt notice any bugs

PR #37, opened 2022-12-13
No syncs in texture setup commands

Yanis42:
Made by @sauraen, everything seems fine on emu (this should save render time, estimated by Sauraen at 2-5%)


sauraen:
If this is just the patch file I put up in Discord, it *generally* works on console but there are a few issues where adjustments were needed (i.e. there are corruptions so syncs need to be manually added or "upgraded" from load to pipe syncs).

Testing this on HLE will be useless--it will ignore syncs. Many LLE emulators may also ignore the syncs.

Yanis42:
yeah I basically made a branch, applied the patch you sent and opened the PR after testing if it worked, I'm gonna try on N64 and Wii VC to see what happens

Yanis42:
Tested on Dolphin and a real N64, here's the results: https://imgur.com/a/MGcQeXE

Skybox and scenes (?) seems broken on N64 (with map select), on Wii VC (Dolphin) the pause menu background is the only issue

sauraen:
That was roughly my experience too. Also, charging a spin attack hangs on N64.

When the contents of a texture (like independent of where it is on the screen) are corrupted, such as the skybox, it's likely somewhere there was a manual load sync added which should have been a pipe sync.

When there is flickering graphics, it's likely there is somewhere where currently there's no sync (because it was removed from the texture loading) which needs a pipe sync.

Yanis42:
turns out there's a missing ``gDPLoadTLUTCmd``, which most likely causes all of the issues I experienced with this (skybox and pause menu map are fixed when adding the command back)

Yanis42:
N64 (and most likely Wii VC) still has issues but on emu it's working properly, since I don't have a loader with an usb port I can't try easily on N64 to see what's wrong

Yanis42:
This PR now targets the dev branch (which already have most of the changes from the original PR), I think we can merge this and fix the hardware issues later and add a comment to say it's broken on N64 (and probably Wii VC)

sauraen:
The whole point of this is to make the game go faster on N64. Please don't make HackerOoT just become permanently broken on real hardware.

Yanis42:
I know it's not ideal but I don't have the tools to debug on a real N64 currently so it would be a temporary "solution" until we finally have a fix, also the dev branch is the perfect place for such things since it's not intended to be stable in the first place

Yanis42:
(also I want to keep HackerOoT working on real hardware as a Wii VC enjoyer, I think people should play on real hardware when they can because it gives a better experience imo)

sauraen:
Okay, well, the current changes in this PR against develop are minimal and an improvement, so they're fine.

I'm a little concerned that nobody is ever going to take the time to debug a bunch of different issues to merge develop back to master.

Yanis42:
the master branch is gonna become a release-only branch so we'll make sure everything works perfectly well when we want to merge dev in master

also if no ones makes a fix I'll do it myself as soon as I can, but that won't happen for months probably

ariahiro64:
i think it should be under a define even if enabled by default

Yanis42:
it's gonna be annoying but I'm gonna ifdef this whole thing so we can merge this

Yanis42:
ok, everything is ifdeffed now, I decided to make a new file for gbi.h, it was the easiest solution tbh

I also moved graphic stuff, including this, to a new file ``config_graphics.h``, should be ready to merge

ariahiro64 (approved):
seems like everything is able to be enabled.
we will need revisit this when f3cdex3 is ready but it im satisfied for now.

Yanis42:
merging this tomorrow

ariahiro64:
tommorow never came lol

Yanis42:
Lmao I went to get some milk and never came back

PR #40, opened 2023-01-03
Fix for debugState having the wrong condition

Yanis42:
This prevented compilation when enabling one of the two options but not both at the same time


Yanis42:
closing this in favor of a bigger PR that will improve a bit how the debug features are controlled/drawn

PR #44, opened 2023-01-16
(Draft) Added controls for most debug features and improved some

Yanis42:
This PR adds controls in ``config_debug_controls.h`` for most debug features, also I improved a bit some of them, for instance I translated everything from the camera debugger, though the display is a bit glitched (but it works properly), I'd like to make further improvements to each features in the future™

The camera debugger has A LOT of controls, that's why I put them at the end of the file otherwise it'd be awful to read imo, also I was lazy and didn't worked on the audio debugger yet, hence the draft state (just wanted to open a PR to get some opinions on things to improve)

Anyway, I hope you like it 👍

### DIFF
--- a/include/config/config_debug_controls.h
+++ b/include/config/config_debug_controls.h
@@ -69,4 +69,295 @@
 // Frame advancing
 #define FA_CONTROL BTN_DUP
 
+// -------------------------------------------
+
+/**** [MAP SELECT CONTROLS] ****/
+// Controller port to use
+#define MAP_SELECT_CONTORLLER_PORT CONTROLLER_PORT_1
+
+// Set to true to use a button combo
+#define MAP_SELECT_BTN_COMBO true
+#define MAP_SELECT_BTN_HOLD_FOR_COMBO BTN_L | BTN_R
+
+// Opening the map select
+#define MAP_SELECT_OPEN BTN_Z
+
+// -------------------------------------------
+
+/**** [INVENTORY EDITOR CONTROLS] ****/
+// Controller port to use
+#define INV_EDITOR_CONTORLLER_PORT CONTROLLER_PORT_1
+
+// Set to true to use a button combo
+#define INV_EDITOR_BTN_COMBO false
+#define INV_EDITOR_BTN_HOLD_FOR_COMBO BTN_Z
+
+// Opening the inventory editor
+#define INV_EDITOR_OPEN BTN_L
+
+// -------------------------------------------
+
+/**** [EVENT EDITOR CONTROLS] ****/
+// Controller port to use
+#define EVENT_EDITOR_CONTORLLER_PORT CONTROLLER_PORT_1
+
+// Set to true to use a button combo
+#define EVENT_EDITOR_BTN_COMBO true
+#define EVENT_EDITOR_BTN_HOLD_FOR_COMBO BTN_L
+
+// Opening the event editor
+#define EVENT_EDITOR_OPEN BTN_CDOWN
+
+// Closing the event editor
+#define EVENT_EDITOR_CLOSE BTN_CLEFT
+
+// Next page
+#define EVENT_EDITOR_NEXT BTN_R
+
+// Previous page
+#define EVENT_EDITOR_PREV BTN_Z
+
+// Change bit value
+#define EVENT_EDITOR_CHANGE_VAL BTN_A
+
+// Move to the bit on the left
+#define EVENT_EDITOR_GO_LEFT BTN_DLEFT
+
+// Move to the bit on the right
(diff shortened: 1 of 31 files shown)


include/config/config_debug_controls.h
```diff
@@ -69,4 +69,295 @@
...
+#define INV_EDITOR_BTN_HOLD_FOR_COMBO BTN_Z
+
+// Opening the inventory editor
+#define INV_EDITOR_OPEN BTN_L
+
+// -------------------------------------------
+
+/**** [EVENT EDITOR CONTROLS] ****/
+// Controller port to use
+#define EVENT_EDITOR_CONTORLLER_PORT CONTROLLER_PORT_1
```
Dragorn421, line 102: "contorller"
also elsewhere

include/config/config_debug_controls.h
```diff
@@ -69,4 +69,295 @@
...
+
+// Set to true to use a button combo
+#define EVENT_EDITOR_BTN_COMBO true
```
Dragorn421, line 105: imo this sounds like it would be a `BTN_` combination, not a boolean

Dragorn421 (commented):
random remarks in passing
(feel free to ignore)

include/config/config_debug_controls.h
Yanis42: I need to cleanup some of names yeah, this should be like ``EVENT_EDITOR_USE_BTN_COMBO``

src/code/game.c
```diff
@@ -168,22 +171,15 @@ void GameState_Draw(GameState* gameState, GraphicsContext* gfxCtx) {
...
-        s32 pad;
-        GfxPrint printer;
-
         GfxPrint_Init(&printer);
         GfxPrint_Open(&printer, newDList);
         AudioDebug_Draw(&printer);
         newDList = GfxPrint_Close(&printer);
         GfxPrint_Destroy(&printer);
-    }
 #endif
```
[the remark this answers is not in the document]
Dragorn421: yes `{}` is just a block. in `if () {}` the `{}` isn't special syntax tied to the if, it just wraps some code under a single block that is the single thing below the `if`

Yanis42:
won't work on this more than that so marking as ready to review I guess, haven't tested since but I merged master and fix the typo

Trueffeloot:
Reviewing this will take a bit longer I'll tell when I'm done

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -10557,6 +10557,13 @@ void Player_UpdateCommon(Player* this, PlayState* play, Input* input) {
 
     sControlInput = input;
 
+    if (IS_DAY) {
+        gSaveContext.dayTime -= gTimeSpeed;
+    } else {
+        // doubled to compensate for time moving twice as fast at night
+        gSaveContext.dayTime -= gTimeSpeed * 2;
+    }
+
```
Trueffeloot, line 10566: `gSaveContext.dayTime` needs to be `gSaveContext.save.dayTime`

src/code/z_debug.c
```diff
@@ -223,15 +226,19 @@ void Regs_UpdateEditor(Input* input) {
                     gRegEditor->dPadInputPrev = dPadInputCur;
                 }
 
+                incrementBy1000 = CHECK_BTN_ALL(input->cur.button, REGS_EDITOR_INCDEC_1000)
+                incrementBy100 = CHECK_BTN_ALL(input->cur.button, REGS_EDITOR_INCDEC_100)
+                incrementBy10 = CHECK_BTN_ALL(input->cur.button, REGS_EDITOR_INCDEC_10)
```
Trueffeloot, line 231: semicolons are missing

Trueffeloot (changes requested):
There were a few compiler errors but nothing big so no problem. I'll be checking for functionality in-game now

Trueffeloot:
I'm currently testing with every debug feature activated to ensure everything at all works so I won't be able to always tell which debugger could cause the respective problem

Issue 1: Texture Flickering in Scenes that use "Box Textures"
image
image
I assume this might have something to do with either the camera debugger or the cs debugger? Will need to see

Update: I commented out the audio debugger. That one's causing a lot of problems it seems and i think it's unfinished

Issue 2: Inventory not Drawing and stuttering sounds 
image
This is because of the audio debugger

Issue 3: Opening up the Map Select collides with opening up the Camera Debugger
Camera Debugger gets activated by pressing L + R and Map Select by pressing L + R + Z. We have to get something new here.

That's probably not it. Didn't test everything, I will continue later on

Yanis002:
closing, too old and too inactive so it'd be better to start over (someday)

PR #46, opened 2023-01-31
[Feature] Widescreen support

Yanis42:
This PR provides support for 16:9, most of the diff comes from things that didn't auto-unwide with that one small change inside z_view, so I had to fix everything (and also shoutouts to kentonm for helping me on this)

I fixed most things but not everything, for instance I didn't looked at each debug features as HackerOoT will probably will get new and better ones in the future, though I fixed the wide chars from GfxPrint (so the map select looks fine), speaking of, I removed the controller 2 check to trigger the map select, so now you just need to enable it and use L+R and press Z anywhere in the game to get into it

I wasn't able to fix properly the minimap arrows, inside dungeons it acts funny and outside it moslty works but sometimes the arrows are misplaced, it seems random and I have no idea why lol, also I wasn't able to fix the background being wide inside 2D areas like most houses and shops, though note that this is only happening when the camera is fixed, otherwise it rescales properly, also I'm not 100% satisfied with the text chars, things that should be perfectly centered aren't really (though it's not that bad)

For z_end_title the DL for the "presented by Nintendo" thing is inside assets and since those aren't commit (yet?) I had to copy pasta the array in the actor file to fix the thing being wide, it will use the assets' one if not in widescreen mode otherwise it will use the fixed one

Other changes are a fix for the inventory editor/event editor ifdefs since it prevents compilation and a translated inv editor and using the boot age macro in the map select as the default age option

Also I used macros to handle the changes instead of adding code, I was thinking this would help with conflicts

Anyway, hope you like it, and huge thanks to kentonm again!

EDIT: forgot to mention that but I made an in-game toggle to switch between normal and wide, hold Z+R then press D-Pad Up to toggle widescreen


Yanis42:
Here's a small imgur album with how it looks in-game: https://imgur.com/a/d9JQdwh

include/config/config_widescreen.h
```diff
@@ -0,0 +1,37 @@
+#ifndef CONFIG_WIDESCREEN_H
+#define CONFIG_WIDESCREEN_H
+
+// This file provides macros to handle switching between 4:3 and 16:9 and fixing issues.
+// It is not meant to be something the users can edit.
```
thecozies, line 5: i'd suggest putting this file just under `/include/widescreen.h` instead in order to not confuse people that this is a config file

thecozies:
return is a tad hard to read
image

hmm idk if its related but saving crashed
image

i'll get some more screen caps, overall looks good on console as far as hud goes

thecozies:
overall i think its worth merging regardless!

Yanis42:
I noticed the crash too, I guess I'll use something unused in savecontext or just remove the thing for now :c 

and for everything else I just tried to stay faithful to the original game, I played a bit on my N64 and I wasn't really bothered by the things not being big enough (but idk I can always increase the size, I just need a small break lmao)

and yeah good idea for the header file, I wasn't really inspired

include/z64save.h
```diff
@@ -289,6 +289,9 @@ typedef struct {
     /* 0x1420 */ s16 worldMapArea;
     /* 0x1422 */ s16 sunsSongState; // controls the effects of suns song
     /* 0x1424 */ s16 healthAccumulator;
+#ifdef ENABLE_WIDESCREEN
+                 u8 isUsingWidescreen; // TODO: doesn't get saved?
```
engineer124, line 293: Saving is a bit of a mess, and not really shiftable yet due to substructs that are used as part of saving:
https://github.com/HackerN64/HackerOoT/blob/af4a250529f363467bc116b81322077b039c4c71/src/code/z_sram.c#L5-L73

Because only the struct `Save` is actually loaded when loading a file:
https://github.com/HackerN64/HackerOoT/blob/af4a250529f363467bc116b81322077b039c4c71/src/code/z_sram.c#L426

There is a PR somewhere in OoT decomp addressing this

Yanis42:
fixed most things, I just couldn't get the text with the item icon working properly and tbh idk how to fix that so I'll let someone else do it lol

also I removed the savecontext variable, saving seems fine now (will add it back when it will be shiftable)

ariahiro64 (approved):
seems like other than prerender shenanigans things work

Yanis42:
(merging this since nothing else came up since the approval)

PR #47, opened 2023-01-31
[Fix] Fixed Visual Studio Code being completely lost

Yanis42:
Inverted the way release mode is working so vscode doesn't get confused and gray out code that shouldn't be disabled


ariahiro64 (approved):
looks good to me

PR #49, opened 2023-02-07
[Improvement] Added a toggle for automatic GI object allocation

Yanis42:
After talking on the decomp discord yesterday I realised forcing the automatic alloc for GI objects isn't great as people might actually want an assert to let them have a better control on the memory, I made an assert before allocating to make sure everything's fine
image


Yanis42:
(note that I forced the crash for testing purposes, I have no idea if this actually works in a real scenario but it should)

Yanis42:
gonna merge this one since it's small (and opened at least 6 months)

PR #50, opened 2023-02-07
[Fix] Fixed Lens of Truth having two circles in 4:3

Yanis42:
noticed this yesterday, turns out I just forgot to use a macro


ariahiro64 (approved):
its such a small change im going to just merge it

PR #51, opened 2023-04-22
Add makedepend

CrashOveride95:
This PR aims to add "makedepend" which automatically generates dependency .d files for `make` upon  compilation
This way, if you change a header that's used by a ton of files, `make` will know you have and recompile the relevant stuff
This PR requires further testing to make sure dependencies are working properly, so it's a draft for now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ OVL_RELOC_FILES := $(shell $(CPP) $(CPPFLAGS) $(SPEC) | grep -o '[^"]*_reloc.o' 
 
 # Automatic dependency files
 # (Only asm_processor dependencies and reloc dependencies are handled for now)
-DEP_FILES := $(O_FILES:.o=.asmproc.d) $(OVL_RELOC_FILES:.o=.d)
+DEP_FILES := $(O_FILES:.o=.d) $(O_FILES:.o=.asmproc.d) $(OVL_RELOC_FILES:.o=.d)
 
 
 TEXTURE_FILES_PNG := $(foreach dir,$(ASSET_BIN_DIRS),$(wildcard $(dir)/*.png))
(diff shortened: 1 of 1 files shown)


ariahiro64 (commented):
tested this on my personal mod and it seems way easier. im going to wait on yandis to merge but im not sure what else needs to be done

CrashOveride95:
I flagged it as draft because I believe reloc file's depend files are overwriting the original .C's depends, so I'd like to make sure nothing's being overwritten before it's pushed

Yanis42:
might sound dumb but will this behave properly once the headers are properly split? it's a work in progress on decomp side (there's some PRs that splits parts of z64.h)

also really cool that we can finally change a header without having to make clean everytime, gotta try this! oh also I'm a bit concerned about the amount of files this will add but idk whatever I guess, we don't have any good solution for this atm so it's better than nothing

CrashOveride95:
This PR doesn't depend on how the headers themselves are organized; if a huge header organization change is made a `make clean` is all that's necessary

Wdym by "amount of files"?

CrashOveride95:
After a brief check everything looks OK, nothing's conflicting, I think this is safe to merge whenever ready

Yanis42:
> This PR doesn't depend on how the headers themselves are organized; if a huge header organization change is made a `make clean` is all that's necessary
> 
> Wdym by "amount of files"?

forgot to reply yesterday sorry, I can phrase it that way: how many files will this add? like one per header or one per C file? or both?

CrashOveride95:
One per C file, each file has a list of all the headers the .c depends on
image

Yanis42:
ahhh it's just text files right, ok then it's totally fine idk why I was thinking it was something heavier lol

gonna try it soon!

ariahiro64 (approved):
been testing this for 2 months and it has no issues other than if you remove a file. however i think this is good to go as is.

PR #53, opened 2023-06-24
Preserve PlayState offsets without ENABLE_FRAMERATE_OPTIONS

HailToDodongo:
Preserves offset of all members of the PlayState struct when ENABLE_FRAMERATE_OPTIONS is disabled.

For context, i am building a map-editor witch needs to access the members of this struct in RAM.
While i can resolve most variables by name using the z64.map file, member inside struct needs to be hard-coded.
It would help a lot if, at least by default, this struct stays the same.


ariahiro64 (approved):

ariahiro64:
such a tiny change ill just merge

FazanaJ:
ramspace preservationists when 8 bytes have just been obliterated from the ecosystem 😲

PR #55, opened 2023-06-26
Restore N64 button colors

recardo-7:
Restore the original button colors from N64 instead of the current GC colors.
Based on [this list](https://pastebin.com/u628WqyU) by Roman.


Yanis42:
can you put those changes under a define inside the config folder? (config_game.h maybe?), I think it would be nice if people could choose and not enforce the thing

otherwise that's a good thing to add :D

recardo-7:
yeah at first I thought why would you want to use gc colors but it's always good to have an option. Added a define in config_game.h

Yanis42 (commented):
code is ok to me but I haven't tried this (@ariahiro64)

Trueffeloot:
@recardo-7 I noticed a few issues while testing

First issue is that the outlines of the cursor are still lighting green on equipment items or songs that require the A button to be pressed:
issue-cursor-outlines

Second issue is that the small box at the bottom of the inventory still has a green A button:
issue-inventory-text
issue-inventory-text2

This a button issue also applies to playing an ocarina song that includes the A Button:
image

Everything else should be good to go

ariahiro64 (changes requested):
Colors should be 100% consistent before its merged.

Yanis42:
I will be taking over this PR since recardo seems not available (and I'd like to merge this), gonna merge master and fix the issues with the pause menu

first time I'm doing this so hopefully it goes well lol

Yanis42:
ok everything should be fixed now

Trueffeloot:
@Yanis42 Seems good from what I can tell. The issues I reported are fixed. This is good to go and the A button in the pause menu can be fixed later 👍

Yanis42 (approved):

PR #56, opened 2023-07-07
Weekly Decomp Merge (07/07/2023)

Yanis42:
we back yay!

note: it's most likely okay but gamestates may shift idk but probably not since it's using enum names now


Yanis42:
@ariahiro64 can you merge that if everything's fine?

PR #57, opened 2023-08-02
Turned z64compress into a git submodule

Yanis42:
The libdeflate submodule is annoying me a lot so I figured I should make the whole thing a submodule, also this will make updating this easier assuming 1. lz4 is cool for oot and 2. it gets merged in z64compress

anyway, let me know if that breaks anything with git, git can be annoying sometimes (also I don't really know git well so hopefully I did nothing wrong)

also I'm not sure if this is a good idea but oh well


Yanis42:
gonna close this for now until I get a better way to implement this without breaking anything

PR #63, opened 2023-09-05
Add Widescreen toggle to SaveContext

Yanis42:
This isn't perfect yet, ideally we would keep the widescreen value outside of the current file but it's fine for now, I'd like to make that a fileselect option and this isn't the point of this change

### DIFF
--- a/include/config/config_safeguards.h
+++ b/include/config/config_safeguards.h
@@ -77,7 +77,7 @@
     #ifdef FORCE_WIDESCREEN
         #define USE_WIDESCREEN true
     #else
-        #define USE_WIDESCREEN (gIsUsingWidescreen == true)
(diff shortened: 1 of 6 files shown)


Trueffeloot:
the button check works.
but uhh
image
It seems like the skybox and other world textures get killed. I'm sure it's not me since I started from a fresh repo.

I have a deku stick on B?
image

If this is because of merging upstream/dev into widescreen_save, it should be fine, else this needs to be fixed

Yanis42:
yeah dev has every pr, including the no sync one which causes those issues

though the stick on B is weird, probably some issue in z_sram.c

Trueffeloot:
in that case, should be fine to be merged since it's anways targeting the dev branch

Yanis42:
the equips should be fixed now

PR #64, opened 2023-09-11
use_kazes_folded_polynomial

ariahiro64:
this uses Kaze Emuar's "folded polynomial" sine and cosine functions. it works by calculating a small section of the polynomial and mirrors it for the rest. this not only saves ram but seems faster than the lookup table.

### DIFF
--- a/include/config/config_game.h
+++ b/include/config/config_game.h
@@ -9,6 +9,9 @@
 // Disables AA (Improves console performance but causes visible seams between unconnected geometry).
 //#define DISABLE_AA
 
+// Use a folded polynomial instead of a lookup table
+#define DISABLE_LOOKUP_TABLE
+
 // Fix annoying glitches (crashes and softlocks)
 #define FIX_ANNOYING_GLITCH
 
--- a/include/macros.h
+++ b/include/macros.h
@@ -214,3 +214,11 @@ extern struct GraphicsContext* __gfxCtx;
     } while (0)
 
 #endif
+
+#ifdef DISABLE_LOOKUP_TABLE
+
+#define ONE 1.0f
+#define SECOND_ORDER_COEFFICIENT 0.0000000010911122665310369f
+#define quasi_cos_2(x) (ONE - SECOND_ORDER_COEFFICIENT * x * x)
+
+#endif
--- a/src/code/z_lib.c
+++ b/src/code/z_lib.c
@@ -1,5 +1,7 @@
 #include "global.h"
 
+#include "config.h"
+
 /**
  * memset: sets `len` bytes to `val` starting at address `dest`.
  *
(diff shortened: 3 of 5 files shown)


include/config/config_game.h
```diff
@@ -9,6 +9,9 @@
 // Disables AA (Improves console performance but causes visible seams between unconnected geometry).
 //#define DISABLE_AA
 
+// Use a folded polynomial instead of a lookup table
+#define DISABLE_LOOKUP_TABLE
```
Yanis42, line 13: I think this name is really confusing, I would name it like ``DISABLE_SIN_COS_LOOKUP_TABLE`` or something like this

include/macros.h
```diff
@@ -214,3 +214,11 @@ extern struct GraphicsContext* __gfxCtx;
     } while (0)
 
 #endif
+
+#ifdef DISABLE_LOOKUP_TABLE
+
+#define ONE 1.0f
```
Yanis42, line 220: why using a macro for this? it looks weird to me 🤔 (the ``ONE`` one)

include/macros.h
```diff
@@ -214,3 +214,11 @@ extern struct GraphicsContext* __gfxCtx;
...
+#define SECOND_ORDER_COEFFICIENT 0.0000000010911122665310369f
```
Yanis42, line 221: where does this number comes from? just out of curiosity

src/code/z_lib.c
```diff
@@ -23,20 +25,56 @@ void Lib_MemSet(u8* dest, size_t len, u8 val) {
     // clang-format on
 }
 
+#ifdef DISABLE_LOOKUP_TABLE
+/**
+ * @param angle binang
+ * @return sinecos(sine, cosine)
+ */
+
```
Yanis42, line 33: that blank space and the one at line 52 bothers me lol (especially the second one)

Yanis42 (commented):
I don't understand a lot how this is working (I'm bad at maths) but here we are

include/config/config_game.h
ariahiro64: done

include/macros.h
ariahiro64: yeah i thought i would need to do the function twice and forgot

src/code/z_lib.c
ariahiro64: done

include/macros.h
ariahiro64: ask kaze i have no idea

Yanis42 (commented):
I'll approve as soon as I test this

include/config/config_game.h
```diff
@@ -9,6 +9,9 @@
 // Disables AA (Improves console performance but causes visible seams between unconnected geometry).
 //#define DISABLE_AA
 
+// Use a folded polynomial instead of a lookup table. This has a speed boots and makes the calculation more percise.
```
Yanis42, line 12: percise

ariahiro64:
should be good for testing

Yanis42:
I already said this on discord but I'm writing it here just for history purposes, the camera is acting weird so I'll wait until this is fixed

include/config/config_game.h
```diff
@@ -9,6 +9,9 @@
 // Disables AA (Improves console performance but causes visible seams between unconnected geometry).
 //#define DISABLE_AA
 
+// Use a folded polynomial instead of a lookup table. This has a speed boots and makes the calculation more precise.
```
Yanis42, line 12: boots

src/libultra/gu/sins.c
```diff
@@ -1,12 +1,17 @@
...
 #include "sintable.inc.c"
-
+#endif
 /**
  * @param angle binang
  * @return sin(angle)*0x7FFF
  */
 s16 sins(u16 angle) {
+#ifdef DISABLE_SIN_COS_LOOKUP_TABLE
+return Math_SinS(angle) * 0x7FFF;
```
Yanis42, line 13: this function isn't declared (just add ``#include "functions.h"``) and it returns a f32, not s16, I feel like we should add a cast here (same comment for coss for the cast)

Yanis42 (changes requested):
the camera is still acting weird

ariahiro64:
farores wind crashes

ariahiro64:
so ive been looking into this
the issues we are having is due to the lookup table snapping where the snapping is causing unhanded behavior based on nintendo just not anticipating it. this also effects controls. we need discussions on whether its actually important to address this or make the user do it as its only slightly harmful to certain camera cases that would require heavy fixes and possible rewrite. this would make syncing difficult for little pay off.

Yanis002:
closing this, it's old and also not 100% functional, some things are breaking and this is getting too inactive, feel free to reopen once it's updated and fixed!

PR #70, opened 2023-10-01
compression rewrite

ariahiro64:
I essentially did 3 things with this commit.

1. Each compression algorithm uses its own function.

2. Added zlib as a compression algorithm as it isnt terrible on a real control deck. This includes a resizable buffer.

3. Every compression algorithm is forced to use Ofast which greatly improves decompression speed

### DIFF
--- a/include/functions.h
+++ b/include/functions.h
@@ -24,10 +24,21 @@ void Main_ThreadEntry(void* arg);
 void Idle_ThreadEntry(void* arg);
 void ViConfig_UpdateVi(u32 black);
 void ViConfig_UpdateBlack(void);
+#ifdef COMPRESSION_YAZ
 void* Yaz0_FirstDMA(void);
 void* Yaz0_NextDMA(u8* curSrcPos);
 void Yaz0_DecompressImpl(Yaz0Header* hdr, u8* dst);
 void Yaz0_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
+#ifdef COMPRESSION_LZO
+void Lzo_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
+#ifdef COMPRESSION_APLIB
+void ApLib_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
+#ifdef COMPRESSION_ZLIB
+void Zlib_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
 void Locale_Init(void);
 void Locale_ResetRegion(void);
 u32 func_80001F48(void);
--- a/spec
+++ b/spec
@@ -29,6 +29,9 @@ beginseg
 #ifdef COMPRESSION_APLIB
     include "build/src/boot/aplib.o"
 #endif
+#ifdef COMPRESSION_ZLIB
+    include "build/src/boot/zlib.o"
+#endif
     include "build/src/boot/z_locale.o"
     include "build/src/boot/assert.o"
     include "build/src/boot/is_debug.o"
--- a/src/boot/lzo.c
+++ b/src/boot/lzo.c
@@ -1,5 +1,8 @@
 /* <z64.me> adapted from lzo1x_d.ch */
 
+#pragma GCC push_options
+#pragma GCC optimize ("Ofast")
+
 /* negative indexing distance */
 #define NINDEX 2
 
@@ -15,9 +18,6 @@ struct decoder {
     unsigned char* buf_limit; /* points to end of scannable area  *
                                * of buf; this prevents yaz parser *
                                * from overflowing                 */
-#if MAJORA
-    unsigned char* dst_end; /* end of decompressed block        */
-#endif
 };
 
 static struct decoder dec;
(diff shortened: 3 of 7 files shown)


include/functions.h
```diff
@@ -24,10 +24,21 @@ void Main_ThreadEntry(void* arg);
...
+#endif
+#ifdef COMPRESSION_LZO
+void Lzo_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
+#ifdef COMPRESSION_APLIB
+void ApLib_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
+#ifdef COMPRESSION_ZLIB
+void Zlib_Decompress(uintptr_t romStart, u8* dst, size_t size);
+#endif
```
Yanis42, line 41: I would use ``#else if (defined XXX)`` there

spec
```diff
@@ -28,6 +28,9 @@ beginseg
 #endif
 #ifdef COMPRESSION_APLIB
     include "build/src/boot/aplib.o"
+#endif
+#ifdef COMPRESSION_ZLIB
+    include "build/src/boot/zlib.o"
```
Yanis42, line 33: same here about ``#else``

src/boot/lzo.c
```diff
@@ -209,10 +209,5 @@ void Yaz0_Decompress(unsigned int pstart /* physical rom offset of file      */
...
     }
 L_done:
-    do {
-    } while (0);
-#if MAJORA
-    dec.dst_end = op;
-    dec.buf_end = 0;
-#endif
-}
+    do { } while (0); }
```
Yanis42, line 212: this do while can be removed right? like it does literally nothing from what I can tell

src/boot/z_std_dma.c
```diff
@@ -377,7 +377,18 @@ void DmaMgr_ProcessRequest(DmaRequest* req) {
...
+#endif
+#ifdef COMPRESSION_LZO
+                Lzo_Decompress(romStart, ram, romSize);
+#endif
+#ifdef COMPRESSION_APLIB
+                ApLib_Decompress(romStart, ram, romSize);
+#endif
+#ifdef COMPRESSION_ZLIB
+                Zlib_Decompress(romStart, ram, romSize);
+#endif
```
Yanis42, line 391: ``#else`` thing like before

src/boot/zlib.c
```diff
@@ -0,0 +1,1290 @@
...
+			&state, sizeof(state)
+		);
+		dst += size;
+		if (!result)
+			break;
+	}
+	
+#if MAJORA
+	dec.buf_end = 0;
+#endif
```
Yanis42, line 1287: since you removed MM stuff from other algorithms you should probably remove it here too for consistency

Yanis42 (changes requested):
I haven't tried it yet but I noticed few things in code

Yanis42:
(also don't use merge soon label when you make the PR, use the waiting for review one instead and when it gets approved use merge soon)

include/functions.h
Yanis42: about this, using else make sure you don't use multiple algorithms at the same time since it doesn't make sense to begin with

ariahiro64:
that should be the requested changes however we need to discuss the default memory size for zlib and perhaps use config.h?

Yanis42:
maybe create a new file called ``config_compression.h`` so we can put compresison stuff in it, otherwise that solution seems good to me

Yanis42 (commented):
code is fine but I need to build this so I'll approve it once that is done

ariahiro64:
in the process of applying changes but keep in mind this isnt tested on wiivc

ariahiro64:
Just checked and Wii VC doesnt work even before this commit. So that will need to be addressed but that is out of the scope of this pr.

include/config/config_compression.h
```diff
@@ -0,0 +1,7 @@
+#ifndef CONFIG_COMPRESSION_H
+#define CONFIG_COMPRESSION_H
+
+/* Size of libdeflates's buffer in kibibytes (yes kib not kb). HackerSM64's implimentation uses a full Mb we use 32 kib*/
+#define KIB 32
+
+#endif
```
Yanis42, line 7: newline missing there (at the end of the file)

Yanis42 (commented):
same as before I need to test before approving

include/config/config_compression.h
Yanis42: also you should probably rename the KIB macro to something more specific to libdeflate, also there's a typo I missed on the comment there ("implimentation")

ariahiro64:
VC should now be fixed. Like stated VC doesn't benefit from optimization in fact it causes crashes. This should also be low maintenance for users. if people need insane compression the can just use zlib that's also in this pr. n64 mode uses the normal optimization levels so its mostly unchanged.

ariahiro64:
Using make wad also forces VC mode as well, I did put a warning to make clean.

Yanis42:
I tried this PR everything seems to work, I requested changes I made and sent as a diff patch so I'm gonna wait until those are applied then I'll proceed with the merge process

Yanis42:
what's going on with this pr lol the diff is suspicious

wouldn't it be better to remake it to target the current dev branch?

ariahiro64:
this is a year old and may be quicker to redo than fix.

PR #75, opened 2023-10-13
Implement Rainbow System

Yanis42:
This is something I did 9 months ago and it was [saved as a gist](https://gist.github.com/Yanis42/146bd6d4c9dec51f71aaa4822a9e4ae4) but here's the HackerOoT implementation!

This adds a global rainbow (see ``main.c`` and ``game.c``) but you can create a private one easily:
- Create a ``Rainbow`` variable (example: ``Rainbow myRainbow``)
- Call ``Rainbow_Init`` in an init function with the address of the variable you just created (example: ``Rainbow_Init(&myRainbow)``
- Call ``Rainbow_Update`` in an update function with the same address (example: ``Rainbow_Update(&myRainbow)``

The default color change speed is set to 8 (frames) but you can change that easily, for instance ``gRainbow.speed = 20;``.

To get the rainbow's color you can do ``gRainbow.color.r``/``gRainbow.color.g``/``gRainbow.color.b``

hackerainbow


Yanis42:
closing as a pr was made for 2.0

PR #78, opened 2023-10-17
Implement Ares GDB

Yanis42:
This is very useful even though idk how to be efficient with it, this was done by HailToDodongo and other people, I only set it up for HackerOoT

Note: on WSL2 you might need to add rules in Windows' Firewall to allow traffic between WSL and Windows (yes, it's dumb) and you will need to make your WSL IP address static, to that iirc you need to write ``nameserver IP`` inside ``/etc/resolv.conf`` where IP is the IPv4 address of the Windows WSL virtual network adapter (you can see that with ``ipconfig /all``), networking is dumb on WSL2 unfortunately 🥲 

As mentioned in the readme of this branch, you will also need one of the latest Ares builds (and might enable ipv4 too in the options), the port used in the launch config should be the default one you can get from Ares

Also this is targeting the main branch instead of the dev one, I was thinking it's the kind of feature we would want asap in main (also I started this on main anyway)


README.md
```diff
@@ -144,6 +144,21 @@ The control mapping on Wii VC will be the following:
 
 Note that you can use the Wii Classic Controller too.
 
+## Using GDB
+### How to setup
+You will need a nightly build of [Ares](https://github.com/ares-emulator/ares)
```
Dragorn421, line 149: meanwhile gdb server made it to an ares release (since ares v134)

README.md
```diff
@@ -144,6 +144,21 @@ The control mapping on Wii VC will be the following:
...
+
+- For Linux/WSL you need to install ``gdb-multiarch`` and [this](https://pypi.org/project/gdb-tools/) Python module, then use the launch configuration ``Ares GDB (Linux)``. If using WSL, you might need to disable the IP change for WSL and update it in the launch configuration, also you might need to add a rule in the firewall to allow traffic. If you're on native Linux simply make the address ``localhost:PORT`` and it should work (port being the one shown on Ares).
```
Dragorn421, line 151: what is https://pypi.org/project/gdb-tools/ for?

Makefile
```diff
@@ -126,20 +126,24 @@ ZAPD       := tools/ZAPD/ZAPD.out
 FADO       := tools/fado/fado.elf
 
 ifeq ($(COMPILER),gcc)
-  OPTFLAGS := -Os -ffast-math -fno-unsafe-math-optimizations
+	ifeq ($(DEBUG_BUILD),1)
+		OPTFLAGS := -Og -g -g3 -ffast-math -fno-unsafe-math-optimizations
```
Dragorn421, line 130: `-g -g3` is the same as `-g3`
`-ggdb` also exists and is supposedly better when using gdb

```suggestion
		OPTFLAGS := -Og -ggdb3 -ffast-math -fno-unsafe-math-optimizations
```
I recommend reading `man gcc` about -g options

Dragorn421 (commented):
Hey I can also shameless-plug my documentation here https://github.com/Dragorn421/z64-romhack-tutorials/blob/master/debugging/index.md

README.md
Yanis42: iirc it was for one of the two python scripts, but idk if they have been updated to remove that requirement since (unless it was never required and I got baited by an unused import)

Dragorn421: afaik it's not needed for either script

Yanis42:
closing this as it's already part of develop/2.0

PR #79, opened 2023-11-09
Fixed typo for slingshot string display list

Javarooster-png:
Changed DL extraction and reference in z_player_lib.c for "gLinkChildSlinghotStringDL" to "gLinkChildSlingshotStringDL"


Yanis42:
This should be opened on normal decomp considering it's not a HackerOoT issue

Javarooster-png:
> This should be opened on normal decomp considering it's not a HackerOoT issue

Alright, will do!

PR #82, opened 2024-02-03
HackerOoT v2.0

Yanis42:
I wanted to do a fresh start on HackerOoT, so basically I started over from decomp/main, which is also what this ``rework`` branch is based on (hence the different commit history), I thought it would be easier to review this way but looks like it's gonna be painful anyway, thankfully I did lots of commits just in case

---
## What are the changes in this branch?
- every feature from config files were implemented again
- much better readability
- HackerOoT is now its own OoT version (like ``gc-eu-mq-dbg`` for instance) though this will need a small change to make it prettier
- like decomp now the output will be located in the build folder (currently ``build/hacker-mq/oot-hacker-mq.z64``)
- now you can toggle HackerOoT features, so basically choosing one of the vanilla versions will disable HackerOoT
- new safeguards to improve readability
- some of the current features were improved, for instance now everything related to ``BOOT_`` defines are implemented inside ``TitleSetup_SetupTitleScreen()``
- probably other stuff I forgot to mention (like more push block speed implementations, I did it for milk crates and the moving wall from forest temple)
- Ares GDB is implemented (.vscode is not commited but there's templates in ``docs/vscode.md``)
- Now you need to supply your own ``common-key.bin`` file instead of generating one automatically when creating wads
- also one important change I forgot to mention, now every config defines are defined and you toggle them with true/false instead of commenting it

### Changes since last time
- now you can run ``make patch`` to generate a BPS patch (not compatible with the compressed rom yet)
- now compressed rom have the algorithms in the output name (example: ``hackeroot-mq-compressed-aplib.z64``), and the ``touch``) command is used to update z_std_dma.c and the spec file, only when compressing
- F3DEX3 has been introduced! when running ``setup`` it will extract the F3DEX2 binaries from the baserom and apply a BPS patch on it to get the F3DEX3 binaries (another reason why I had to introduce flips), this branch has the required changes and some of the recommendations, I don't understand a lot this so I'll leave the rest of the implementation to someone else
- I moved installation instructions outside the main readme
- I added a "contributors" list in the readme, with every person who got a merged commit here
- I added links to the tools we introduced to HackerOoT (gzinject, F3DEX3, z64compress and Flips in this branch)

## Is there anything missing?
- the "no anti-aliasing" feature isn't back yet (though I disabled the pause menu AA), F3DEX3 has been implemented now but I still didn't look at this again, soon™
- the "make depend" feature isn't there yet, I'll probably add it back but with a makefile toggle as it can be annoying sometimes (for instance when you change the scene table it's rebuilding every assets, which is bad)
- probably other stuff I just forgot but will come back once I go through the merged PRs

---
Let me know if there's any issue/question/whatever, I said this a lot but I'm opened to suggestions so feel free!

Also I had to do stupid things in python scripts to make the custom version work properly, this will be improved once decomp handles that properly


Yanis42:
new changes arrived, see ``Changes since last time`` in this PR's description for the list of the new things
just gonna do the things from ``Is there anything missing?`` and it should be ready

Yanis42:
I'm gonna merge this, I just have two things to add but I'm not very motivated to work on that for now so I'll just keep it pinned somewhere to remind me I need to do it

the rework branch will become main at some point, for now I'll keep both (rework being a secondary branch for now, might change that in few months) and I'll keep pushing decomp updates to this branch, it builds properly on my computer so hopefully it's ok

PR #84, opened 2024-04-14
[2.0] Implemented Motion Blur

Yanis42:
same as #45 except I made a small improvement, now you can disable motion blur and keep your motion blur commands in your cutscenes


Yanis42:
I'll merge this in the next days if nothing else happens here

PR #88, opened 2024-04-20
Avoid disabling no syncs in texture load

Yanis42:
this is a bit broken currently, also this enables the map select by default (I think it's a nice thing to have enabled)


Yanis42:
I'm gonna merge this now because this is important

PR #94, opened 2024-04-21
Widescreen toggle

recardo-7:
Add a toggle for widescreen in the start menu


Yanis42:
I know this is a draft but my first suggestion would be using the mod_assets folder since it's there now, instead of commiting these files, second I just wanna say we talked about how it would be nice to have a page 2 in the options of the file select, not saying you should do it of course, I have that on my todo list

Reonu:
Yeah definitely commit the assets to the mod_assets folder. Files placed in the normal assets folder will be deleted after a make distclean even if they were commited into git.

ariahiro64:
i think it could be on the side like my character select in cg
however my code is shitty and should be rewritten if used in hackeroot
already removed it temporarily anyway

PR #96, opened 2024-04-22
Add GPU timings detection (gGPUTimingsExist)

Reonu:


README.md
```diff
@@ -67,3 +67,4 @@ List of every HackerOoT contributors, from most recent to oldest contribution:
 - ariahiro64
 - ghost
 - krm01
+- Reonu
```
Yanis42, line 70: you should move your name at the top of the list since it's ordered from recent to oldest, alternatively you can reverse the order of that list

Yanis42 (commented):
nice, just one tiny thing

README.md
krm01: dont reverse it pls i wanna be at the bottom

PR #102, opened 2024-04-25
Add Gamecube support

Yanis42:
started as a joke but tbh it's kinda cool, not finished yet as the emulator part isn't ready yet so this will be a draft (also this PR is fixing an issue on Mac OS with gzinject and it's fixing a small issue with sym_info.py)


Yanis42:
finally fully working! huge thanks to cadmic who helped me making it work (turns out the issue was libultra, for now we're compiling it with IDO but only when we target the GameCube)

PR #103, opened 2024-05-05
Better noclip

krm01:
move with joystick and c-buttons


Yanis42:
(also small reminder for next time, the base branch need to be develop/[latest] (so 2.1.0 in this case))

PR #105, opened 2024-05-08
60 Hz on PAL consoles with correct audio

Reonu:
This PR makes it so that HackerOoT outputs 60 Hz even on PAL consoles, but keeps the code that compensates for the audio timings being slightly different on said consoles. So you get 60 Hz with the correct audio pitch.

### DIFF
--- a/include/variables.h
+++ b/include/variables.h
@@ -44,6 +44,7 @@ extern OSViMode osViModeMpalLan1;
 extern OSViContext* __osViCurr;
 extern OSViContext* __osViNext;
 extern OSViMode osViModeFpalLan1;
+extern OSViMode gCustomViModePal60Lan1;
 extern u32 __additional_scanline;
 extern const char gBuildMakeOption[];
 extern const char gBuildGitVersion[];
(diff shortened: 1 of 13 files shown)


Yanis42 (approved):
can't say if it works as intended but it builds properly and it seems to boot properly on ntsc

Reonu:
> can't say if it works as intended but it builds properly and it seems to boot properly on ntsc

For clarity, I've tested it on my PAL N64 and it works.

If you're using a flashcart that reports region accurately (like the summercart), you get 60 Hz with the correct audio pitch correction for PAL systems.

If you're using a flashcart that lies about region and reports NTSC always, like the Super64, you get 60 Hz with the audio having a slightly wrong pitch, just like before. But nothing can be done about this, since the flashcart is lying about the region, so we can't detect that it's a PAL system.

Yanis42:
let me know whenever it's ok to merge this

Reonu:
You can merge this now.

Thanks to Tharo's change, PAL consoles will now output a more correct 60 Hz with the correct timings, which will also be compatible with more devices. In other words, PAL consoles will now simply output 60 Hz with virtually no drawbacks.

A 50 Hz toggle might still be worth it for certain terrible setups, but this can be explored further as it adds far more complexity for very little benefit, so I don't think it needs to be in this PR.

PR #107, opened 2024-05-09
(Draft) New Inventory Editor

Yanis42:
this PR change how the inventory debug/editor works, instead of having a single menu we're navigating the pause menu and press buttons to give/take away/increase amounts of items, including the health, double defense, magic meters, and amount of rupees, small keys, health, magic and you can even change the dungeon items for every dungeon (boss keys, maps and compasses)

everything is implemented already as this is a port of something I already had working on older HackerOoT, though there's lots of small issues here and there that I need to fix, hence the draft state for now, also there's features I'd like to add, like giving dungeon items from the map screen for instance

also you can edit the deku stick/nut capacity tier and other things like that (though the display is glitched atm I think, not sure why yet)

feel free to give feedback or give suggestions


ariahiro64:
im surprised this isnt merged. the oot one no longer works and i really need it for my sanity.

Yanis42:
this isn't merged because there's issues that needs to be fixed and I lack motivation, see the top comment from ``inventory_editor.c``

Yanis002:
iirc someone fixed the issues but idr who but feel free to ping me again about it so we can get this merged

Yanis002:
made some progress today, only things left are magic bar issues to address and also another minor drawing issue that happens on the equipment screen when showing the misc upgrades then going back to the normal ones, will merge this PR once those are fixed

PR #112, opened 2024-05-12
Multiple F3DEX3 versions

sauraen:
Adds support for multiple F3DEX3 versions, with only one in RAM at a time and the ability to load different ones in from ROM. There are eight total versions: profiling default, A, B, and C, and for each base or NOC (no occlusion planes). Options in the debug menu are provided to switch microcodes.

On the build side, no `make f3dex3` is needed anymore: for most users, they will just build as usual, and the Makefile will take care of extracting F3DZEX2, patching all the different versions, and converting them to object files to be included in the ROM. The only F3DEX3 related manual build action is `make create_f3dex3_patches`, which will be run by maintainers to convert a new F3DEX3 version to the bps patches.

This also fixes some extremely minor oversights in a few places in the codebase, and slightly cleans up the debug menu logic (not changing the logic).


data/rsp.rodata.f3dex3.s
Yanis42, line 1: I don't think we need this file anymore

Yanis42 (commented):
nice work!

data/rsp.rodata.f3dex3.s
sauraen: We do, or at least it would have to be reorganized. This currently contains three other non-F3D-family microcodes (audio, S2D, jpeg). On my personal repo I moved all the microcodes--those three plus rspboot--into a single file. Do you want me to do that here?

Yanis42: sorry forgot to reply, doesn't ``rsp.rodata.s`` already contain those? ``rsp.rodata.f3dex3.s`` was copy pasted from that one

sauraen: Well sure of course that file does, but that file also contains F3DZEX2, so that file can't be included in a F3DEX3 build. Hence the duplication. I was offering to change it to be one file with all the common microcodes and then an optional file with F3DZEX2.

Yanis42: ooooh right I see, I think one file would be better yeah

Yanis42 (approved):

PR #115, opened 2024-05-26
Update F3DEX3 version to resolve issues

sauraen:
- Adds checking for imminent overflow in clipped polygon generation and abort clipping (fixes crash when standing on firewood pile in Kakariko)
- Fixes overlay 4 commands which use cmd_w1_dram (fixes incorrect branch W behavior)
- Removes clipping triangulation with minimal scanlines algorithm (see EX3 readme)


Yanis42 (approved):
seems to work properly

PR #116, opened 2024-05-27
New profiler with RSP and CPU tracing, replacing vanilla SPEEDMETER

sauraen:
This PR replaces the vanilla SPEEDMETER profiler, which has some issues, with a much more capable optional profiler under the debug menu. Its UI modes are:
- Hidden
- Real FPS display
- Virtual FPS display (1 / graphics time)
- Graphics profiling
- Graphics tracing
- CPU profiling
- CPU tracing
- All tracing

The graphics profiling mode shows RDP counters and CPU-measured RSP times, plus performance counters from the currently selected version of F3DEX3 if enabled. The tracing modes show traces (time bars) on screen of when each RSP task and CPU thread was running, allowing easy identification of scheduling problems.


Yanis42 (approved):
really nice work! can't say I understand everything but I looked at the code to see if there was anything I could say about but it seems fine to me, it builds properly and it seems to work as intended

PR #118, opened 2024-05-27
Clean up F3DEX3 support

sauraen:
- Removes the separate defines for enabling different F3DEX3 features (except for `ENABLE_F3DEX3_NOSYNCS` because that is broken). All "RECOMMENDATIONS", "LIGHT_RECO", and "NEW_FEATURES" are always present if you have F3DEX3 enabled.
- Fixes the broken implementation of the F3DEX3 lighting changes.
- Adds specular light sizes--they're not really great as is, but they're at least vaguely reasonable. If they're left as zero, specular won't work.
- Adds `SPCameraWorld` support, which is required for specular and Fresnel.

This finishes all the game engine changes for F3DEX3 except for the occlusion planes system, which is coming next.

### DIFF
--- a/include/config/config_graphics.h
+++ b/include/config/config_graphics.h
@@ -7,15 +7,17 @@
 
 /**
  * F3DEX3 options
- * This only works on "LLE-Emulators" like ares or ParaLLEl. It will not work on standard Project64.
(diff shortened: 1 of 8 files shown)


Yanis42 (approved):
builds properly, boots properly, and seems to work as expected according to my little scene

PR #126, opened 2024-05-31
Use ``s64`` instead of ``s16`` for actor parameters

Yanis002:
this was an idea I had after talking with @Reonu about how actor parameters works last month

I kept the pointers because I wanted to make sure it worked fine but imo they're all completely useless, also I used a signed long long instead of an unsigned because it was raising warnings, I compared with current develop/2.1.0 and I have the same number of warnings now so everything should be fine hopefully (ngl I didn't tested this extensively but overall it seems to work properly)

so with this change, if I'm not dumb, instead of having parameters on ``0xABCD``, we can have parameters on ``0xABCDEFGHIJKLMNOP``

### DIFF
--- a/include/z64actor.h
+++ b/include/z64actor.h
@@ -258,7 +258,7 @@ typedef struct Actor {
     /* 0x003 */ s8 room; // Room number the actor is in. -1 denotes that the actor won't despawn on a room change
     /* 0x004 */ u32 flags; // Flags used for various purposes
     /* 0x008 */ PosRot home; // Initial position/rotation when spawned. Can be used for other purposes
-    /* 0x01C */ s16 params; // Configurable variable set by the actor's spawn data; original name: "args_data"
+    /* 0x01C */ s64 params; // Configurable variable set by the actor's spawn data; original name: "args_data"
     /* 0x01E */ s8 objectSlot; // Object slot (in ObjectContext) corresponding to the actor's object; original name: "bank"
     /* 0x01F */ s8 targetMode; // Controls how far the actor can be targeted from and how far it can stay locked on
     /* 0x020 */ u16 sfx; // SFX ID to play. Sfx plays when value is set, then is cleared the following update cycle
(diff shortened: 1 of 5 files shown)


Thar0:
Increasing the width of the params field is not so easy as just changing some types unfortunately. Some actors depend on 16-bit sign extension for correct behavior (look for just about anywhere that checks `params < 0`). Further, 64-bit types do not compile to particularly nice code under the O32 ABI; 32-bit params are surely sufficient for most situations, with the usual z-rotation fallbacks in particularly pathological cases?

I think this needs some careful discussion, it's a topic that doesn't seem like it should be complex but I think it may be.

Yanis42:
ok well, I made a it a draft for now I guess

Yanis42:
> 32-bit params are surely sufficient for most situations, with the usual z-rotation fallbacks in particularly pathological cases?
tbh I'm always really frustrated by ``s16``s when it comes to parameters, I like to add tons of features to the actors I make or edit so it gets full quickly and I have to make sacrifice or use the rotations, which I find weird btw

one solution would be using a secondary parameter variable, I assume this would solve most concerns here? not sure how it would affect compilation though (thinking about ``ActorEntry``

Reonu:
> > 32-bit params are surely sufficient for most situations, with the usual z-rotation fallbacks in particularly pathological cases?
> 
> tbh I'm always really frustrated by `s16`s when it comes to parameters, I like to add tons of features to the actors I make or edit so it gets full quickly and I have to make sacrifice or use the rotations, which I find weird btw
> 
> one solution would be using a secondary parameter variable, I assume this would solve most concerns here? not sure how it would affect compilation though (thinking about `ActorEntry`

having a second variable sounds way more reasonable and safe to me, it's less likely to break vanilla behavior and it makes it easy to nicely separate vanilla behavior from custom behavior when editing vanilla actors

Yanis002:
closing for now

PR #139, opened 2024-06-17
fix pause screen from being occluded

ariahiro64:
just a simple check if ithe game is paused. that prevents so may bugs with occlusion planes.


Yanis42:
(make sure to run ``format.py``)

ariahiro64:
did it anyway bit it is complaining about things that werent even my fault lol

Yanis42:
you need to use clang-format-14 iirc

ariahiro64:
fedora doesnt have 14

ariahiro64:
this kind of sucks because such a small thing is now blocking me from contributing

ariahiro64:
all checks pass now. thank distro box

Yanis42 (commented):
it works but the side effect is it's not occluding at all once you pause, and it start occluding again when you unpause

I'm not sure if that's fine or not (@sauraen)

sauraen:
This is okay and does solve this problem, but given that the occlusion plane also can cull the A button, I figured it was better to make a more comprehensive fix rather than just disabling it while paused: https://github.com/HackerN64/HackerOoT/pull/140

PR #147, opened 2024-07-29
Merge ``zeldaret/oot``'s main

Yanis42:
works on my machine™


Yanis42:
the audio stuff is done (except a clean up PR I think) so this is finally good to go, merging in 24 hours from now

PR #154, opened 2024-11-17
HackerOoT clean-up

Yanis002:
WIP clean-up, opening this as a draft so people can follow the steps (and force me to work on it), feel free to make suggestions!


Reonu:
If 1.2 is the chosen baserom then IMO there should be an define to disable the censorship introduced in 1.2 (Fire Temple song and textures), even if it requires the user to provide both 1.2 and 1.0 baseroms.

Yanis002:
yeah good point, I don't remember what I wanted to do for the cleanup tbh but I'll try to review this and see what I can add

Reonu:
The crash when using the profiler on other ROMs should be fixed before merging, but other than that, I guess the censorship stuff can wait, better to merge now and fix that latter than to let this PR rot

Yanis002:
alright, this should be ready now I think, I don't have anything else in mind for now and this cleanup is kinda important

PR #156, opened 2024-12-02
Add Hiisuya's New Actor Script

recardo-7:
All credit to @hiisuya 

Python script to streamline the process of adding a new actor

### DIFF
--- a/new_actor.py
+++ b/new_actor.py
@@ -1,0 +1,262 @@
+import os
+import re
+import sys
+import shutil
+import argparse
+import textwrap
+from os import path
+from pathlib import Path
+
+createObject = True
+useModAssets = False
+useActorProfile = False
+
+def convertActorName(name: str):
+    actorSpec = "_".join(it[0].upper() + it[1:] for it in name.split("_"))
+    actorDefine = "ACTOR_" + actorSpec.upper()
+    objectSpec = "object" + "_" + name.lower()
+    objectDefine = objectSpec.upper()
+    actorFileName = "z_" + name.lower()
+
+    return dict(
+        actorSpec = actorSpec,
+        actorDefine = actorDefine,
+        objectSpec = objectSpec,
+        objectDefine = objectDefine,
+        actorFileName = actorFileName
+    )
+
+# check if folders exists, if not, create them and necessary files
+def makeFilesAndFolders(folderName, fileName, objectName):
+    filePathActor = Path(path.curdir, "src/overlays/actors", f"ovl_{folderName}")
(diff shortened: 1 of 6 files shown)


new_actor.py
```diff
@@ -0,0 +1,262 @@
+import os
```
Yanis002, line 1:
```suggestion
#!/usr/bin/env python3

import os
```
this allows people to do ``./new_actor.py`` instead of ``python3 new_actor.py``

template_files/object.h
```diff
@@ -0,0 +1,4 @@
+#ifndef {objectSpecCap}_H
+#define {objectSpecCap}_H 1
```
Yanis002, line 2:
```suggestion
#define {objectSpecCap}_H
```
not consistent with the other header, tbh I don't think setting a value is that necessary

new_actor.py
```diff
@@ -0,0 +1,262 @@
...
+    actorFileLine = addToTables(names["actorSpec"], names["actorDefine"], names["objectSpec"], names["objectDefine"])
+    addToSpec(names["actorSpec"], names["actorFileName"], names["objectSpec"], actorFileLine)
+    completeFiles(names["actorSpec"], names["actorDefine"], names["actorFileName"], names["objectSpec"], names["objectDefine"])
+
+    print(f'{names["actorSpec"]} created!')
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-name", required = True, help = "Name of actor. Use alphanumeric and/or _ as valid characters", default = "")
+    parser.add_argument("--noobject", required = False, help = "Optional. Use if you do not want to create an object for this actor", action='store_false')
```
Yanis002, line 237:
```suggestion
    parser.add_argument("-n", "--name", required = True, help = "Name of actor. Use alphanumeric and/or _ as valid characters", default = "")
    parser.add_argument("--noobject", required = False, help = "Optional. Use if you do not want to create an object for this actor", action='store_false')
```
not a fan of the inconsistency between ``-name`` and ``--noobject``

I'm not sure how people are using this but I'd invert the ``--noobject`` option so you could do ``./new_actor.py -n test -o`` to tell you want the object (I can't find a short name for "no object")

Yanis002 (changes requested):
I didn't read deeply the python script but it works so lgtm, we can always improve it later, that said there's few changes I'd like to have

also I was thinking it might be better in the tools folder, especially the template files, though the script seems to specifically check for the root folder, don't bother with this though I'll add it to my cleanup plan™

Yanis002 (approved):

PR #161, opened 2025-01-03
Actor Cutscene system from Majora's Mask

Yanis002:
This ports actor cutscenes back to OoT, also I added more things that were required:
- the cutscene manager (z_eventmgr.c in MM decomp)
- the cutscene camera logic (cutscene_camera.c in MM decomp)
- the letterbox system
- a state flag to halt actors during cutscenes
- some bits of the camera logic
- some math functions that I don't wanna try to understand lol
- missing HUD visibility features

draft until the other PR is merged (since I based this branch on the other one and I don't want to play with git to fix that)


Yanis002:
important things to note if you want to use this:
- the cutscene manager doesn't handle starting a cutscene through the `additionalCsId` field
- to play a cutscene you have to add it to the queue and play it before the next time `Play_Main` calls `CutsceneManager_Update`. This is important as this function clears the "next cutscene queue", and the "waiting queue" is cleared after each playstate update

for reference, here's a debug thing I used to test this (I put that at the end of `Play_Main`):
```c
    // get the additional cutscene id and use it if the value is not -1
    s16 optCsId = CutsceneManager_GetAdditionalCsId(0);
    s16 csId = optCsId >= 0 ? optCsId : 0;

    // check if the cutscene is the next on the queue, if it is play it, otherwise add it to the queue when the button L is pressed
    if (CutsceneManager_IsNext(csId)) {
        CutsceneManager_Start(csId, &GET_PLAYER(this)->actor);
    } else if (CHECK_BTN_ALL(this->state.input[0].press.button, BTN_L)) {
        CutsceneManager_Queue(csId);
    }

    // update the cutscene manager
    CutsceneManager_Update();
    CutsceneManager_ClearWaiting();
```

PR #165, opened 2025-01-07
Title Cards from MM

Yanis002:
This backports title cards from Majora's Mask and adds more customization possibilities, instead of having a text id in the scene table you can add a scene command instead, this is not mandatory and instead it's possible to not use the scene command and it will default to a generic data (if a title card is used of course, which is defined by the entrance table iirc)

I made generic names for vanilla scenes but I don't like it much tbh, it's not really shiftable and I don't think I can achieve that (or maybe by generating the data with a python script? meh)

note: the gradient texture was made by me so it should be safe to commit


Yanis002:
I removed what I didn't like, instead of using placeholder names I made a default entry named "Undefined" for vanilla scenes, idk what to do and it's dumb not to merge this because of that, will merge tomorrow if nothing comes up (and CI is happy)

PR #170, opened 2025-03-21
Add ACTOR_FLAG_FREEZE_EXCEPTION support

Zelllll:


Yanis002 (approved):
seems to work! I'm pushing a fix to the cutscene manager here since it's related, will merge when CI is happy

PR #171, opened 2025-03-30
Merge zeldaret/main

Yanis002:


Yanis002:
This PR is no longer a draft, which means it will get merged soon if there isn't any comment about it (builds & boots properly), this brings important changes so I'll copy paste what I said on discord: 

"PSA: few things changed in decomp (which will end up in HackerOoT in the near future), first the "segmented_to_virtual" macro was renamed "os physical to k0", if you used that in your mod it might break, then `global.h`, `macros.h` and `variables.h` are gone (and soon `z64.h` too but it's still there for assets compatibility purposes), that means you'll have to clean up your mod if you want to update HackerOoT with the latest changes (which aren't there yet), basically the headers are split now

I'm not sure how I should handle that for HackerOoT but I followed decomp's steps in cleaning up the headers, idk what people want exactly for modding but feel free to make suggestions (as always) :muscle: "

PR #176, opened 2025-07-04
Fix LF normalization causing build issues

Zelllll:


Yanis002 (approved):
CRLF is the dumbest thing every created

PR #187, opened 2025-11-08
Bugfixes and New MM options

NekoNeko6330:
I've added 3 new config options and fixed errors for the Camera Debugger

AUDIO_DELAY: Delay the audio update by 3ms (causes the audio to go out intermittently)

CLAMPED_JUMP_VELOCITY: Clamps the jumping velocity (Disable for uncapped jumping speed like in MM!)

MM_BOTTLE_MODEL: Use the Majora's Mask Model for Bottles


include/config/config_game.h
```diff
@@ -26,6 +29,9 @@
 // Enables the beep sound at low health (vanilla behavior)
 #define ENABLE_LOW_HEALTH_BEEP true
 
+// Clamps the jumping velocity (vanilla behavior)
```
Yanis002: I think the comment should mention disabling this is giving jumps the MM behavior (except the animation I guess)

include/config/config_game.h
```diff
@@ -12,6 +12,9 @@
 // Add a fix for the Stale Reference Manipulation glitch (OoT's Pandora's Box)
 #define FIX_SRM true
 
+// Delay the audio update by 3ms (causes the audio to go out intermittently)
+#define AUDIO_DELAY false
```
Yanis002: where did you get the audio cutting bug? I know it happens on Wii VC/GameCube but are there PC emulators affected too?

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -7171,11 +7171,11 @@ void func_8083DF68(Player* this, f32 arg1, s16 arg2) {
 
 void func_8083DFE0(Player* this, f32* arg1, s16* arg2) {
     s16 yawDiff = this->yaw - *arg2;
-
+#if CLAMPED_JUMP_VELOCITY
```
Yanis002: just a detail but can you add empty lines before the #if and after the #endif?

src/code/z_message.c
Yanis002: I think this file should be reverted since #188 has the proper fixes for title cards (sorry for the wait on that btw)

src/code/z_demo.c
```diff
@@ -1,5 +1,6 @@
 #pragma increment_block_number "gc-eu:128 gc-eu-mq:128 gc-jp:128 gc-jp-ce:128 gc-jp-mq:128 gc-us:128 gc-us-mq:128" \
                                "ique-cn:128 ntsc-1.0:0 ntsc-1.1:0 ntsc-1.2:0 pal-1.0:0 pal-1.1:0"
+#include "padmgr.h"
```
Yanis002: is it needed? or is it a leftover? no worries either way it's just an include but I'm curious

src/code/audio_thread_manager.c
```diff
@@ -27,7 +27,7 @@ void AudioMgr_NotifyTaskDone(AudioMgr* audioMgr) {
  */
 void AudioMgr_HandleRetrace(AudioMgr* audioMgr) {
     AudioTask* rspTask;
-
+#if AUDIO_DELAY
```
Yanis002: would be nice to have empty lines before and after here too :pray:

include/config/config_graphics.h
```diff
@@ -56,7 +59,7 @@
  * Use the button combo Z + R + D-Pad Up to toggle
  * Note: pre-rendered scenes with the fixed camera mode won't re-scale properly
 */
-#define ENABLE_WIDESCREEN false
+#define ENABLE_WIDESCREEN true
```
Yanis002: I feel like this should be treated like `ENABLE_F3DEX3` so people can turn it on or off in the makefile or elsewhere on the system, I personally don't use widescreen since I'm only testing and implementing stuff so I would be bothered a lot if this was enabled by default, but on the other hand I understand why people would want it so I basically just want a compromise :muscle:

assets/objects/object_link_boy/bottle_contents.i8.png
Yanis002: this and the other texture are what will be blocking the merge, I can't allow it since it's from MM and I want to avoid committing assets (basically mirroring normal decomp rules), I don't know how this works but one idea I had was using BPS patches over the OoT textures or something, or have a script that extracts the necessary assets from an MM baserom because I also wanted to commit more stuff using MM textures and couldn't because of that (the wizzrobe from the actor pack I believe)

this is an ongoing issue for a while and it need addressing but idk what would be the best way, I'm open to ideas

Yanis002 (changes requested):
sorry lot of stuff, nice work!

include/config/config_game.h
NekoNeko6330: Every emulator I've tested on, have reported this in the discord months ago

include/config/config_graphics.h
NekoNeko6330: ENABLE_WIDESCREEN just allows you to enable widescreen by pressing Z+R+D-UP, it doesn't force widescreen by default, is that an issue?

Yanis002: yeah I'm just dumb lol, I forgot you had to enable it in-game

Yanis002 (changes requested):
this is better thanks, can you also add the DL stuff to `object_link_child`? I think it's not working for child link currently

PR #190, opened 2025-11-15
Add Sauraen's improved anim morphing

Reonu:
Adds Sauraen's improved animation morphing. Here's an example of the difference it makes.

Before (vanilla): 

https://github.com/user-attachments/assets/e2eb9fe0-9641-4279-b12f-76a2c035b07e

After:

https://github.com/user-attachments/assets/bb94ca66-cae1-43ec-ad01-0873322d3054

This is pretty much necessary for custom skeletons to be usable

### DIFF
--- a/src/code/z_bettermorph.c
+++ b/src/code/z_bettermorph.c
@@ -1,8 +1,17 @@
-// See https://github.com/sauraen/OoTAnimInterp for more general information about OoT's math conventions
-// and the design of this patch. Algorithms are modified from:
-// https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
-// http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/
-// http://www.euclideanspace.com/maths/geometry/rotations/conversions/quaternionToEuler/Quaternions.pdf
+/**
+ * @file z_bettermorph.c
+ *
+ * This file contains an implementation of a more accurate animation interpolation routine. This avoids common
+ * issues encountered when linearly interpolating Euler angles during animations, instead using quaternions to
+ * perform spherical interpolation.
+ *
+ * See https://github.com/sauraen/OoTAnimInterp for the original implementation, particularly the README for more
+ * general information about OoT's math conventions and the design of this routine.
+ * Algorithms are modified from:
+ *  - https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
+ *  - http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/
+ *  - http://www.euclideanspace.com/maths/geometry/rotations/conversions/quaternionToEuler/Quaternions.pdf
+ */
 
 #include "config.h"
 #include "ultra64.h"


include/config/config_graphics.h
```diff
@@ -54,6 +54,13 @@
 */
 #define ENABLE_MOTION_BLUR true
 
+/*
+ * Improved animation morphing by Sauraen (see copyright notice in ther repo readme)
```
Yanis002: typo?

README.md
```diff
@@ -95,3 +95,24 @@ List of every HackerOoT contributors, from most recent to oldest contribution:
...
+```
+Copyright (C) 2021 Sauraen
+You may redistribute, create derivative works of, and otherwise copy this
+software, subject to the following conditions:
+- If the software or modified versions of it are distributed in source form,
+  it must retain this copyright notice and list of conditions
+- Sauraen must be credited (e.g. in the Special Thanks) of any project using
+  this software or modified versions of it
+
+This software does not contain any information derived from the Gigaleak.
```
Thar0: This kind of note need not be reproduced here. It's self-evident by the fact it's even considered for inclusion into this repo

README.md
```diff
@@ -95,3 +95,24 @@ List of every HackerOoT contributors, from most recent to oldest contribution:
...
+
+See README.md for more general information about OoT's math conventions and the
+design of this patch. Algorithms are modified from:
+https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
+http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/
+http://www.euclideanspace.com/maths/geometry/rotations/conversions/quaternionToEuler/Quaternions.pdf
```
Thar0: This refers to the original repo's readme, so either link to there or reword/remove this

README.md
```diff
@@ -95,3 +95,24 @@ List of every HackerOoT contributors, from most recent to oldest contribution:
...
+This repository contains sauraen's improved animation morphing code, which can be found [here](https://github.com/sauraen/OoTAnimInterp/blob/master/animinterp.c). The following is the copyright notice pertaining to said code:
+
+```
+Copyright (C) 2021 Sauraen
+You may redistribute, create derivative works of, and otherwise copy this
+software, subject to the following conditions:
+- If the software or modified versions of it are distributed in source form,
+  it must retain this copyright notice and list of conditions
+- Sauraen must be credited (e.g. in the Special Thanks) of any project using
+  this software or modified versions of it
```
Thar0: This condition is highly unusual as far as open source licenses go. The closest analogue I can think of is the zlib license's similar clause:
> 1. The origin of this software must not be misrepresented; you must not
   claim that you wrote the original software. If you use this software
   in a product, an acknowledgment in the product documentation would be
   appreciated but is not required.

The zlib license particularly singles out project documentation as the appropriate place to include an acknowledgement of the covered software. Hre however it's a little unclear to what extent the credit condition applies. Does it apply to in-game credits sequences, open source repo READMEs, notices bundled with release patches, or all of these? If a user enabled this feature in HackerOoT, what obligations does that impose? If I clone HackerOoT, enable this feature, and ship a patch, that doesn't seem like I'm really providing adequate credit for this feature.

cc @sauraen for clarification on the intended meaning

Thar0 (changes requested):

README.md
Thar0: This is not a part of the copyright notice

README.md
Reonu: Fixed. Thanks

Reonu:
I fixed co-authoring

README.md
Yanis002: worst case scenario I'll just mention where the `IMPROVED_ANIMATION_MORPHING` define is the fact that enabling this feature means the original author must be credited or something like that

README.md
```diff
@@ -83,6 +83,7 @@ List of every HackerOoT contributors, from most recent to oldest contribution:
 
 - hiisuya
 - Zeldaboy14
+- Sauraen
```
krm01: this list is supposed to be ordered from newest to oldest btw

Yanis002: yeah I corrected that, sauraen contributed before and was never credited here that's why the location might seem odd

sauraen:
Hey folks. Glad my code was useful.

At the time I wrote that "license" I was not anticipating that the code would get used in a community library-style project with many contributors which would then be reused by many other people. The model I was expecting was: one person makes one romhack, copies code from various places on the internet, and fails to credit their sources.

You have my permission to include the animation code directly in HackerOoT without any special license terms different from the rest of HackerOoT.

Yanis002 (approved):

src/code/z_bettermorph.c
```diff
@@ -0,0 +1,155 @@
+// See https://github.com/sauraen/OoTAnimInterp for more general information about OoT's math conventions
+// and the design of this patch. Algorithms are modified from:
+// https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
+// http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/
+// http://www.euclideanspace.com/maths/geometry/rotations/conversions/quaternionToEuler/Quaternions.pdf
```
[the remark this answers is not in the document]
Yanis002: thanks

Thar0 (approved):

PR #194, opened 2025-11-27
spec changes

Arthurpandaroux:
Some changes to the spec/linkerscript that improves cache performance for the grf.


Arthurpandaroux:
uh it looks like some changes from my last commit are in it. My bad

Arthurpandaroux:
I messed up so i'm closing it for now I got mixed up with my different branches

Issue #195, opened 2025-11-29
Can't make changes to Soundfont_X.xml files without editing the extracted folder


Reonu:
There's currently no way to add SFXs without editing the extracted folder because you can't edit the Soundfont_X.xml files otherwise. Copying them to the regular assets folder and editing that produces this error:

<img width="1264" height="152" alt="Image" src="https://github.com/user-attachments/assets/798297e4-cf79-47e4-81d6-8e4d8dffc00e" />

PR #196, opened 2025-11-30
new_sfx.py script to easily add SFX samples to audio decomp

krm01:
i'm  hoping someone will be willing to take this and test / adapt to hackerOOT use. this is a script i wrote for indigo. it will edit files and wire things up to give you an enum that can be called in code as a oneshot, like `Sfx_PlaySfxCentered(NA_SE_EV_CUSTOM_FOO);`.

this also extends the environmentbank to hold 512 samples instead of 256, and then all the custom samples go there. an alternative implementation would create a whole new sample bank just for custom stuff. go ahead if you want to do that, i couldn't be bothered lol.

oh, this also relies on `ffmpeg` being available in the path as an implicit dependency to convert the .wav file to proper format. added sfx's always seem to quiet so this wrapper has an optional param that boosts the volume before converting and makes things much more better.

usage:

```sh
./new_sfx.py --volume 5 My-Custom-SFX.wav
```
example output:

image

usage help is available from the script by passing `--help`:

image

### DIFF
--- a/new_sfx.py
+++ b/new_sfx.py
@@ -1,0 +1,272 @@
+#!/usr/bin/env python3
+"""
+Adds a new sound effect to the project by processing a .wav file,
+updating necessary XML / build files and creating an enum that can
+be called in game. Covers 80% of use cases where you just want to
+add a new SFX and play it as a oneshot in code. Taken from Indigo
+project, this script can be freely used/shared without attribution
+"""
+
+import os
+import re
+import sys
+import shutil
+import argparse
+import logging
+import subprocess
+import textwrap
+import xml.etree.ElementTree
+import xml.dom.minidom
+from pathlib import Path
+
+
+LOG: logging.Logger             = logging.getLogger(f"[{Path(__file__).name}]")
+LOG_LEVEL: int                  = logging.INFO
+
+EXTRACTED_PATH: Path            = Path("extracted")
+SAMPLEBANK_NAME: str            = "SampleBank_0"
+ASSETS_SAMPLES_PATH: Path       = Path("assets") / "audio" / "samples" / SAMPLEBANK_NAME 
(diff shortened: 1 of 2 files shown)


Yanis002:
is this (partially?) addressing #195? I was planning on making a workaround until decomp fixes it so that's a nice timing actually (the idea would be moving the xml to edit to the root assets folder and have something to delete extracted files on `make setup`)

btw it's fine if it's not fixing the issue I mentioned, I'm just curious

krm01:
umm i think the script will install the permanent stuff into assets/ so it can be committed if you haven't already done that as part of preparing the directories and files, so yea i think so?

oh, one thing that probably does need tweaking actually, i just assume the first "game" it finds in the version list is the one to use, im guessing you'd want that to be changed to be the hackeroot game or w/e.

new_sfx.py
```diff
@@ -0,0 +1,272 @@
...
+    os.makedirs(ASSETS_SAMPLES_PATH, exist_ok=True)
+    os.makedirs(ASSETS_SAMPLEBANK_XML.parent, exist_ok=True)
+    os.makedirs(ASSETS_SOUNDFONT_XML.parent, exist_ok=True)
+
+    games: list[str] = os.listdir(EXTRACTED_PATH)
+    if len(games) == 0:
+        LOG.error("No assets found in extracted/, 'make setup' not yet run?")
+        sys.exit(1)
+    # just use the first one found, it's fine (probably)
+    game = games[0]
```
krm01, line 85: this is what i am referring to in my previous comment above. prolly wanna edit this part to be hackeroot-y

Yanis002: yeah I think having a `--version -v` argument would be convenient for HackerOoT, people don't necessarily use ntsc-1.2 (which is the default currently)

krm01: right. Oh or probably reading it from the Makefile would be better

Yanis002: I think both would be neat actually, that said I don't think anyone would use several versions at the same time lol

